PR #5, opened 2023-11-02
organize codes directory sturcture

karim20230:
organize directory structure holds scripts and configurations for automating  provisioning, and management of infrastructure


henrirosten (changes requested):
If you need to split the `.tf` file, you need to add proper license and copyright header to all those new files. For reference, see: https://github.com/tiiuae/ghaf-infra/blob/d55ed4b2d1d7b7b24de33130472c67b052db1aea/terraform/azure-ghaf-infra.tf#L1C1-L3

PR #7, opened 2023-11-03
Create full-func vm

karim20230:
- Define Azure Resource Group for Ghaf Hydra Infrastructure
-  Create Virtual Network (VNet) with Address Space 10.0.0.0/16 
- Establish Subnet within the VNet with Address Prefix 10.0.2.0/24 
- Set Up Availability Set for High Availability
- Implement Network Security Group with SSH Access Rule 
- Allocate Public IP for Access
- Implement Network Security Group with SSH Access Rule
 - Provision Storage Account for Data Persistence
- Deploy Virtual Machine with Ubuntu 22.04 LTS

### DIFF
--- a/terraform/azure-ghaf-infra.tf
+++ b/terraform/azure-ghaf-infra.tf
@@ -1,36 +1,57 @@
 # SPDX-FileCopyrightText: 2023 Technology Innovation Institute (TII)
 #
 # SPDX-License-Identifier: Apache-2.0
-
 # Resource group
-resource "azurerm_resource_group" "rg" {
-  name     = "ghaf-infra-terraform-dev"
+resource "azurerm_resource_group" "ghaf_infra_tf_dev" {
+  name     = "ghaf-infra-tf-dev"
   location = var.resource_group_location
 }
-# Create  VN
-resource "azurerm_virtual_network" "ghaf-infra-vnet" {
-  name                = "ghaf-infra-terraform-dev-vnet"
-  address_space       = ["10.3.0.0/24"]
-  location            = azurerm_resource_group.rg.location
-  resource_group_name = azurerm_resource_group.rg.name
+# Create VN
+resource "azurerm_virtual_network" "ghaf_infra_tf_vnet" {
+  name                = "ghaf-infra-tf-vnet"
+  address_space       = ["10.0.0.0/16"]
+  location            = var.resource_group_location
+  resource_group_name = azurerm_resource_group.ghaf_infra_tf_dev.name
 }
-
-
-# Create public IPs
-resource "azurerm_public_ip" "ghafhydra_terraform_public_ip" {
-  name                = "ghaf-infra-terraform-dev-ip"
-  location            = azurerm_resource_group.rg.location
-  resource_group_name = azurerm_resource_group.rg.name
+# Create Subnet
+resource "azurerm_subnet" "ghaf_infra_tf_subnet" {
+  name                 = "ghaf-infra-tf-subnet"
+  resource_group_name  = azurerm_resource_group.ghaf_infra_tf_dev.name
+  virtual_network_name = azurerm_virtual_network.ghaf_infra_tf_vnet.name
+  address_prefixes     = ["10.0.2.0/24"]
+}
+# Network interface
+resource "azurerm_network_interface" "ghaf_infra_tf_network_interface" {
+  name                = "ghaf-infratf286-z1"
+  location            = var.resource_group_location
+  resource_group_name = azurerm_resource_group.ghaf_infra_tf_dev.name
+  ip_configuration {
+    name                          = "my_nic_configuration"
+    subnet_id                     = azurerm_subnet.ghaf_infra_tf_subnet.id
+    private_ip_address_allocation = "Dynamic"
+    public_ip_address_id          = azurerm_public_ip.ghaf_infra_tf_public_ip.id
+  }
+}
+# Create Availability Set
+resource "azurerm_availability_set" "ghaf_infra_tf_availability_set" {
+  name                         = "ghaf-infra-tf-availability-set"
+  location                     = var.resource_group_location
+  resource_group_name          = azurerm_resource_group.ghaf_infra_tf_dev.name
+  platform_fault_domain_count  = 2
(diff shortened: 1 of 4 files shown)


terraform/azure-ghaf-infra.tf
```diff
@@ -44,3 +82,55 @@ resource "azurerm_network_security_group" "ghafhydra_terraform_nsg" {
...
+  os_profile {
+    computer_name  = "ghafinfratf"
+    admin_username = "karim"
+
+    custom_data = <<-EOF
+              #!/bin/sh
+              echo 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDAKcuBPeS7BJqdTDsKvXbEmO13NV0A+cjpI4TNORsMbB2lxT9n8A/aqSuj4IpPnYBtQrS21Y5n1gfCvHAGe059iSVERS6tUP+D1goPjkDwqlclfmRWCzosdi5cC+k1g87dNTtDhsxPLxoYEgqyiCJxB1cha+GEn7uIGo+/GC9+hKC/ryOl2he20yFVahvymnPYbQ9+Zvsk6LJlmeuSgVBoiTLKx8OFRI7vBEr3+pHuZgyXYSVhO/F58x1tA3x8VxW075ZVCEudAwmKwtXA8uGuxx64V6sg7CX6i+E+SML/TIpXMH0Uq5IF9/rjkC6KYn4q1kHUij4PxYUFkjYX/BlR5OXTTAg00XdurwoPTlKPLnUNFpetK/270SXOojscnoC8w7s4xu4QqlO093UR3oNcMjAKB3tTWbhlKibTheSbUcqHhZBsULuiS8xlLCnKQy1YnhhOMtezFHzjzM14ptmb2MFnw4j5zbBsHw0TUtQzAAQqio+TpaRQnTk7EDoxAY8=' >> /home/karim/.ssh/authorized_keys
+              sed -i 's/PasswordAuthentication yes/PasswordAuthentication no/' /etc/ssh/sshd_config
+              systemctl restart sshd
+              EOF
```
henrirosten, line 129: This will probably not work.
If the command is executed as root, `/home/karim/.ssh/authorized_keys` will be owned by root which is obviously not correct.

If the command is not executed as root, the sed command to edit the /etc/ssh/sshd_config would fail.

henrirosten: Also, the sed replacement is a bit fragile. For instance, what if there are many spaces or tabs before 'yes' ?

henrirosten (changes requested):

terraform/azure-ghaf-infra.tf
henrirosten: If there is no better way of doing this in terraform,
perhaps we could completely re-write `/etc/ssh/sshd_config` in the intial VM config?

Once the NixOS config gets installed, the ssh configuration will be overwritten and maintained by the nix flakes configs in this repo.

karim20230:
addressed the comments and verified with Terraform apply

terraform/azure-ghaf-infra.tf
```diff
@@ -43,4 +64,62 @@ resource "azurerm_network_security_group" "ghafhydra_terraform_nsg" {
...
+    publisher = "canonical"
+    offer     = "0001-com-ubuntu-server-jammy"
+    sku       = "22_04-lts-gen2"
+    version   = "latest"
+  }
+  admin_username                  = "karim"
+  disable_password_authentication = true
+  admin_ssh_key {
+    username   = "karim"
+    public_key = file("~/.ssh/id_rsa_nixos.pub")
```
henrirosten, line 101: You are assuming local file `~/.ssh/id_rsa_nixos.pub` is available. I get a validation error from this line, since the file `~/.ssh/id_rsa_nixos.pub` is not available on my local system.

You should not assume `~/.ssh/id_rsa_nixos.pub` is available, but provide the value of the public key inline.

terraform/azure-ghaf-infra.tf
```diff
@@ -43,4 +64,62 @@ resource "azurerm_network_security_group" "ghafhydra_terraform_nsg" {
...
+  name                 = "customScript"
+  virtual_machine_id   = azurerm_linux_virtual_machine.ghafinfra_tf.id
+  publisher            = "Microsoft.Azure.Extensions"
+  type                 = "CustomScript"
+  type_handler_version = "2.1"
+  settings = jsonencode({
+    commandToExecute = <<-SCRIPT
+      #!/bin/bash
+      sudo apt-get update
+      sudo apt-get install -y apache2
```
henrirosten, line 115: Why do we need to install apache2 here?

terraform/azure-ghaf-infra.tf
```diff
@@ -43,4 +64,62 @@ resource "azurerm_network_security_group" "ghafhydra_terraform_nsg" {
...
 }
-
+# Create Storage Account
+resource "azurerm_storage_account" "ghafinfra_tf_storage_account" {
+  name                     = "ghafinfrastorage"
+  location                 = var.resource_group_location
+  resource_group_name      = azurerm_resource_group.ghaf_infra_tf_dev.name
+  account_tier             = "Standard"
+  account_replication_type = "LRS"
+}
```
henrirosten, line 74: Isn't the storage account already created in https://github.com/tiiuae/ghaf-infra/blob/766b4fb424668cdc64c90c88101fcb72516f4114/terraform/azure-storage/tfstate-storage.tf#L32 ?

henrirosten (changes requested):

terraform/azure-ghaf-infra.tf
henrirosten: I think we should completely remove the creation of the storage account resource here.

terraform/azure-ghaf-infra.tf
henrirosten: I think we should configure the admin user public key here, as well as disable the password authentication like you already do. Then we can remove the 'customScript' resource completely.

karim20230: it's validated and and able to do the ssh

henrirosten (approved):

PR #9, opened 2023-11-06
Pin nixos-anywhere, fix README

henrirosten:
- Pin nixos-anywhere to prevent possible issues caused by the upstream PR: https://github.com/nix-community/nixos-anywhere/pull/243. Read more from the nixpkgs issue: https://github.com/NixOS/nixpkgs/issues/265014.
- Fix the task.py usage instructions in README.md after https://github.com/tiiuae/ghaf-infra/pull/8.
- Make `nix flake check` pre-push check less verbose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,50 +48,72 @@ Run the following command to list the available tasks:
 $ invoke --list
 Available tasks:
 
-  build-local         Build NixOS configuration `target` locally.
-  deploy              Deploy NixOS configuration `target` to host `hostname`.
-  install             Install `target` on `hostname` using nixos-anywhere, deploying host private key.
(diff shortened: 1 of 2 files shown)


cazfi:
Usually changes unrelated to each other should not go to the same commit (they can be in the same PR), but I'm not requesting splitting this commit now that it would take some effort. Just a note for the future.

joinemm (approved):

henrirosten:
> Usually changes unrelated to each other should not go to the same commit (they can be in the same PR), but I'm not requesting splitting this commit now that it would take some effort. Just a note for the future.

I agree, but we are still far from production (or using this tooling anywhere), so I think it's fine for now until we declare it's not :)

PR #14, opened 2023-11-10
Initial terraform config for azure arm VM

henrirosten:
- PoC azure arm VM configuration with terraform
- Refactor `azure-ghaf-infra.tf` to sections per VM-specific resources. This is the first step to make it easier to add new VMs.
- Move the terraform resource group location to 'northeurope' due to the needed arm VM resources not being available in 'swedencentral'
- Add `.direnv` directory to `.gitignore`

### DIFF
--- a/terraform/azure-ghaf-infra.tf
+++ b/terraform/azure-ghaf-infra.tf
@@ -11,6 +11,13 @@ terraform {
       source = "carlpett/sops"
     }
   }
+  # Backend for storing tfstate (see ./azure-storage)
+  backend "azurerm" {
+    resource_group_name  = "ghaf-infra-storage"
+    storage_account_name = "ghafinfrastatestorage"
+    container_name       = "ghaf-infra-tfstate-container"
+    key                  = "ghaf-infra.tfstate"
+  }
 }
 provider "azurerm" {
   features {}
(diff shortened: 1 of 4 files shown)


terraform/azure-ghaf-infra.tf
```diff
@@ -19,19 +26,10 @@ provider "azurerm" {
...
-    storage_account_name = "ghafinfrastatestorage"
-    container_name       = "ghaf-infra-tfstate-container"
-    key                  = "ghaf-infra.tfstate"
-  }
-}
 # Resource group
 resource "azurerm_resource_group" "ghaf_infra_tf_dev" {
   name     = "ghaf-infra-tf-dev"
-  location = "swedencentral"
+  location = "northeurope"
```
karim20230, line 32: Nice, we are moving to more reliable, av zones! 

Btw also maybe would nice if we parameterize the variable:

```
variable "location" {
description = "The Azure region where resources will be created."
type = string
default = "northeurope"
}

```

henrirosten: I would propose we parameterize later when we actually have a use-case that requires the resource group location (or other options) to be configurable. 

Might be my lack of Terraform experience, but right now, I don't see the benefit or use-case for adding the extra variable.

@karim20230 : feel free correct me if I'm wrong, or fail to see the use-case.

karim20230: @henrirosten  yessir  we can do it in future ..

karim20230 (approved):

PR #15, opened 2023-11-10
Configuration for binarycache vm

joinemm:
Add configuration for nixos binary cache (new cache.vedenemo.dev) that runs in a qemu vm on ficolo. it exposes nix-serve-ng through http (https soon when we get the dns records changed).

includes disko config to format the second disk and mount it under `/nix` to hold the nix store

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -75,6 +75,7 @@ TARGETS = OrderedDict(
     {
         "build01-dev": TargetHost(hostname="51.12.57.124", nixosconfig="build01"),
         "ghafhydra-dev": TargetHost(hostname="51.12.56.79", nixosconfig="ghafhydra"),
+        "binarycache-ficolo": TargetHost(hostname="172.18.20.109", nixosconfig="binarycache"),
     }
 )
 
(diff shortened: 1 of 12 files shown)


tasks.py
```diff
@@ -75,6 +75,7 @@ class TargetHost:
     {
         "build01-dev": TargetHost(hostname="51.12.57.124", nixosconfig="build01"),
         "ghafhydra-dev": TargetHost(hostname="51.12.56.79", nixosconfig="ghafhydra"),
+        "binarycache": TargetHost(hostname="172.18.20.109", nixosconfig="binarycache"),
```
henrirosten, line 78: Could we name the 'binarycache' alias to something that would indicate the environment it's going to be used in. Maybe 'binarycache-ficolo'?

henrirosten:
@joinemm : can you rebase the commits on top of current main to get rid of the merge commit in the history? Also, consider squashing the commits unless you want to show the intermediate changes.

henrirosten (approved):

PR #17, opened 2023-11-13
Add ubuntu aarch64 builder configuration

henrirosten:
- Add a shell script that configures aarch64 ubuntu host so that it can be used part of the buildfarm. Run the script on applying the terraform configuration.
- Start collecting some common terraform errors and solutions to the terraform README.md.
- Apply formatting fixes to qemu-common.nix and tasks.py.

### DIFF
--- a/terraform/scripts/ubuntu-builder.sh
+++ b/terraform/scripts/ubuntu-builder.sh
@@ -53,8 +53,14 @@ configure_builder () {
     chown -R nix:nix /home/nix/.ssh
     chmod 700 /home/nix/.ssh
     chmod 600 /home/nix/.ssh/authorized_keys
-    # Extra nix config for the builder
+    # Extra nix config for the builder,
+    # for detailed description of each of the below options see:
+    # https://nixos.org/manual/nix/stable/command-ref/conf-file
     extra_nix_conf="
+# 20 GB (20*1024*1024*1024)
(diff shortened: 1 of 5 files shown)


terraform/scripts/ubuntu-builder.sh
```diff
@@ -0,0 +1,148 @@
...
+        sudo ln -fs "$f" "/usr/bin/$(basename "$f")"
+    done
+}
+
+configure_builder () {
+    # Add user: nix
+    sudo useradd -m -d /home/nix nix
+    # Allow ssh login with buildfarm public key
+    mkdir -p /home/nix/.ssh
+    echo 'ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIIQwSH1R9xShZg1w5dZjRagYLae0QFBPYT3i80iHW1Ej' >> /home/nix/.ssh/authorized_keys
```
karim20230, line 52: Maybe sops?

henrirosten: I don't see any reason why we could not have the public key in the repo in clear text. Moreover, decrypting the value from sops would add unnecessary complexity, which I would like to avoid.

karim20230 (approved):

karim20230 (approved):

PR #26, opened 2023-11-22
Add initial monitoring server configuration

joinemm:
Adds monitoring VM running Grafana and Prometheus, that monitors build4 and binarycache using node_exporter metrics

### DIFF
--- a/.sops.yaml
+++ b/.sops.yaml
@@ -10,6 +10,7 @@ keys:
   - &karim age122lvqyrdqz30fkfututykl0yle9u63u2em6e4aut7e5draws83ns3npt3a
   - &jrautiola age15jq5gjjd7ypsdlqfjtqy4red57v8ggqq9na6u3xffznu678nydpsuuwjg0
   - &binarycache age1s47a3y44j695gemcl0kqgjlxxvaa50de9s69jy2l6vc8xtmk5pcskhpknl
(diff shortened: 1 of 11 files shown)


henrirosten (approved):
Approved, but I have not tested this.
@joinemm :
I assume you get a warning on installing the new monitoring configuration, since it doesn't have `secrets.yaml`, so the host key decryption should fail: https://github.com/tiiuae/ghaf-infra/blob/68318451901453382a21321316eb6d8e534986a3/tasks.py#L277

joinemm:
@henrirosten I actually never saw this warning when testing, I was not aware secrets were required

henrirosten:
> @henrirosten I actually never saw this warning when testing, I was not aware secrets were required

It's strictly not required, but if there's no `ssh_host_ed25519_key` for the given nixosconfig, `install` task should issue a warning and ask confirmation from the user. If that's not what happens, then it's probably an error in tasks.py.

joinemm:
added secrets and tested with `install` task

henrirosten:
You are missing a license for the new secrets.yaml, you should add secrets.yaml.license similarly to other host configs:

> WARNING  'reuse lint':
> MISSING COPYRIGHT AND LICENSING INFORMATION
> 
> The following files have no copyright and licensing information:
> * hosts/monitoring/secrets.yaml

henrirosten (approved):

PR #34, opened 2023-12-07
feat: support nixosConfigurations in perSystem

brianmcgee:
Adds a flake.parts module which transposes `nixosConfigurations` defined within `perSystem` functions.

Makes it easier to re-use the `perSystem` `pkgs` instance within `lib.nixosSystem`.

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -4,7 +4,6 @@
 {
   self,
   inputs,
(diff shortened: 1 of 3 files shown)


flokli:
I don't think this is (still?) necessary. Closing, we can still resurrect if needed.

PR #35, opened 2023-12-07
Azure images

flokli:
- disk creation has been moved to use the nixpkgs tooling to build images and boot up machines with the desired NixOS config directly, rather than infecting them post-first-boot
 - new terraform directory/state in "terraform/jenkins"
 - ssh config moved to cloud-init. Changing this still requires VM recreation, but no new image
 - new terraform module to create image and upload into Azure
 - new terraform module to create VM in Azure, exposing knobs for machine identity, portforwards and data disk attachment
 - new machine "binary-cache" defined, exposing a blob storage container / bucket via https
    - uses machine identity to get access to binary cache bucket
 - data disk handling overhauled, longer-living state is now moved to separate data disks, which are mounted by cloud-init


terraform/jenkins/jenkins_controller.tf
```diff
@@ -0,0 +1,102 @@
...
+
+  virtual_machine_name         = "ghaf-jenkins-controller"
+  virtual_machine_size         = "Standard_D1_v2"
+  virtual_machine_source_image = module.jenkins_controller_image.image_id
+
+  virtual_machine_custom_data = join("\n", ["#cloud-config", yamlencode({
+    users = [
+      for user in toset(["bmg", "flokli", "hrosten"]) : {
+        name                = user
+        sudo                = "ALL=(ALL) NOPASSWD:ALL"
```
alextserepov, line 36: Debatable, but I would rather disable SSH password auth (keep it only key based) and require passwords for sudo. That would require setting up some initial password to let user define proper password on first login, but will limit the security risks in case of compromised account.

terraform/jenkins/binary_cache.tf
```diff
@@ -0,0 +1,115 @@
...
+  location            = azurerm_resource_group.default.location
+
+  security_rule {
+    name                       = "AllowSSHHTTPSInbound"
+    priority                   = 400
+    direction                  = "Inbound"
+    access                     = "Allow"
+    protocol                   = "Tcp"
+    source_port_range          = "*"
+    destination_port_ranges    = [22, 80, 443]
```
alextserepov, line 86: Debatable in non m2m world, but since it is internal system, do we need to keep 80 open? It redirects, yes, but what if redirect is misconfigured intentionally or non-intentionally on specific host?

alextserepov (commented):
A bit hard to follow without having a full big picture in mind, but some comments from my side.

terraform/jenkins/jenkins_controller.tf
flokli: This is the current behaviour, wheel doesn't need password: https://github.com/search?q=repo%3Atiiuae%2Fghaf-infra%20wheelneedspassword&type=code

These machines are cattle, all user state and interactively-chosen passwords are wiped whenever the machine is recreated (when we boot a new image or change the user data).

We could potentially track a per user password hash here, and require that password to be entered for sudo, but it's a lot of churn.

Let's discuss this in a followup. We mirror the current behaviour, and I'd rather solve this problem more elegantly with something like SSO for ssh, where we can require an arbitrary set of MFA at the identity provider.

terraform/jenkins/binary_cache.tf
flokli: If we stick with caddy as a reverse proxy, we can probably drop port 80 here. nginx would have needed port 80 for HTTP-01 challenge, but caddy can also do TLS-ALPN-01.

henrirosten: @alextserepov : Notice: nginx is still used as reverse proxy on some other hosts, so we would also have to change to caddy in other host configurations if we want to completely drop port 80.

terraform/jenkins/jenkins_controller.tf
alextserepov: Yeah, I agree we have made some mistakes in the past and I understand the convenience vs security debate, which is much older than myself. :) 
Yes, let's take it up on a followup. 
(SSO sounds a bit like an overkill for this, but let's see).

terraform/jenkins/binary_cache.tf
alextserepov: Like I said, this one is debatable in a non m2m world. As a security guy, I believe that all open-text protocols/ports should be disabled/closed on ingress. Keeping it open/served is asking for trouble, especially if there are/might be open-text-credentials flying. And the price we have to pay with nginx is just to type https:// the first time we open the link in a browser. After that it will keep on choosing HTTPs for this address in the same browser. Not a big deal, taking into account that CI/CD is internal. But I'm happy to stand corrected here. 100% open-minded.

@henrirosten Yes, you are correct. We should unify the system, thus if we switch to caddy here, we will have to switch on other hosts as well. 

Actually, I haven't looked into caddy myself, but I think it is a good idea to research this option. Thanks, @flokli!

flokli:
There's been a bit of back and forth w.r.t. ensuring the data disks get mounted reliably, so I didn't update the branch. Sorry for the radio silence. It's now sorted out.

The config is a shift from the previous method, yes - machines are considered cattle and short-lived. Whenever we change the Nix config of the host, we bake a new *image*, tear down the VM, and start it up again. Only the state on the explicitly created data volumes is preserved (things like ssl certificates).

This means there's not really a need to SSH in and change certain things, or activate a new system configuration, removing the necessity for a lot of the steps documented in the documentation.

Ideally, bare-metal machines would behave very similar - we could have a "write once on every boot, during boot" system partition (or netboot the kernel initrd and run from there), and we'd only need a data partitions where/if we need state.

Yes, the docs need updating, but we're currently in this transition state where we only have the config for some hosts (and I don't want to delete documentation on how to deal with the other currently present hosts).

It might make more sense to do this once we have the full POC to show.

henrirosten (approved):

flokli:
Ok, this should be ready for another review, as most of the nix-specific build infrastructure components are tied together.

 - I had to spend quite some time getting the azure platform support in nixpkgs in shape, so we can mount data volumes (to keep state where needed in an image-based deployment model) and be able to use the resource disk as an overlay for `/nix/store` (with that mount having to happen quite early), which is somewhat required in Azure too, as the system disk is too small to the /nix/store, at least on a builder. It all worked out, but nixpkgs currently points to a version containing these low-level fixes. Once that's merged, we can move to upstream again. 
 -  There's now 2 build machines allowing ssh (only) for a `remote-build` ssh user with that public key. No individual user accounts or secrets live on these machines. (The VM have read-only access to the bucket to substitute from there directly, but that's instance-identity-based and mostly to avoid having to pipe all of this through the `binary-cache` VM, which is more for non-Azure consumers.)
 - There's two secrets used on the jenkins-controller VM (only), an ed25519 ssh key and nix cache signing key. They are provided via Azure Key Vault. We currently use a python script to export these secrets into two files, but I'd like to switch this to [systemd service credentials](https://systemd.io/CREDENTIALS/) and an azure key vault compatible systemd service credentials daemon, so we can use iptables to restrict access to the metadata server to uid 0 only (and `rclone-http.service`, or use another workaround there). This is to prevent other users on the system from accessing the Azure Key Vault as well.
 - `/etc/nix/machines` on the `jenkins-controller` VM is populated with the list of remote builders and their host keys. Currently, this happens once on startup, though we might want to make this more flexible and dynamic in the future. This might require some more machinery, to dynamically register machines and publish host keys when ready.
 - The `jenkins-controller` VM has a post-build-hook configured, which uploads every build into the Azure Storage account, signing with the cache signing key while doing so.

All these machines are dispensable, and except for the few data volumes we mount, loose all state whenever the configuration is changed. Most "environment-specific config" (like hostnames to use) is populated on startup through cloud-init, treating the individual image more like an "appliance" (Nix substituters and binary cache public keys being one annoying exception, but there's also ways to workaround this).

What's missing is configuring the Jenkins pipelines themselves, so Nix builds are triggered - though the setup can already be tested by ssh-ing to the jenkins-controller VM as an individual user (which is untrusted in terms of Nix), and triggering a build of something that's not already in one of the caches.

hosts/jenkins-controller/configuration.nix
```diff
@@ -0,0 +1,179 @@
...
+      EnvironmentFile = "/var/lib/rclone-http/env";
+    };
+  };
+
+  # Configure Nix to use this as a substitutor, and the public key used for signing.
+  nix.settings.trusted-public-keys = [
+    "ghaf-jenkins:5OXpzoevBwH4sBR0S0HaIQCik2adrOrGawIXO+WADCk="
+  ];
+  nix.settings.substituters = [
+    "http://localhost:8080"
```
alextserepov, line 164: https?

hosts/jenkins-controller/get_secret.py
```diff
@@ -0,0 +1,24 @@
+"""
+This script retrieves a secret specified in $SECRET_NAME
+from an Azure Key Vault in $KEY_VAULT_NAME
+and prints it to stdout.
```
alextserepov, line 4: requested clarification in slack.

terraform/jenkins/binary_cache_signing.tf
```diff
@@ -0,0 +1,47 @@
...
+  depends_on = [
+    azurerm_key_vault_access_policy.binary_cache_signing_key_terraform
+  ]
+}
+
+resource "azurerm_key_vault_access_policy" "binary_cache_signing_key_terraform" {
+  key_vault_id = azurerm_key_vault.binary_cache_signing_key.id
+  tenant_id    = data.azurerm_client_config.current.tenant_id
+  # "TerraformAdminsGHAFInfra" group
+  object_id = "f80c2488-2301-4de8-89d6-4954b77f453e"
```
alextserepov, line 40: can we avoid hardcoded object-ids?

terraform/jenkins/binary_cache_signing.tf
```diff
@@ -0,0 +1,47 @@
...
+
+  secret_permissions = [
+    "Get",
+    "List",
+    "Set"
```
alextserepov, line 45: Need to understand big picture on this a bit more. Don't feel comfortable with List & Set. But it depends on the context, naturally.

terraform/jenkins/binary_cache_signing.tf
```diff
@@ -0,0 +1,47 @@
+# nix-store --generate-binary-cache-key foo secret-key public-key
+# terraform import secret_resource.binary_cache_signing_key "$(< ./secret-key)"
+resource "secret_resource" "binary_cache_signing_key" {
+  lifecycle {
+    prevent_destroy = true
```
alextserepov, line 5: what is our resource lifecycle policy in general? Do we have it documented somewhere?

hosts/jenkins-controller/configuration.nix
```diff
@@ -0,0 +1,179 @@
...
+  lib,
+  ...
+}: let
+  post-build-hook = pkgs.writeScript "upload" ''
+    set -eu
+    set -f # disable globbing
+    export IFS=' '
+
+    echo "Uploading paths" $OUT_PATHS
+    exec nix --extra-experimental-features nix-command copy --to 'http://localhost:8080?secret-key=/etc/secrets/nix-signing-key&compression=zstd' $OUT_PATHS
```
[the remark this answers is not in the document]
flokli: This is the post-build-hook from https://nixos.org/manual/nix/stable/advanced-topics/post-build-hook#implementing-the-build-hook. It copies build results (collected from the builders) into the binary cache, signing them while doing so.

hosts/builder/configuration.nix
```diff
@@ -0,0 +1,53 @@
...
+      EnvironmentFile = "/var/lib/rclone-http/env";
+    };
+  };
+
+  # Configure Nix to use this as a substitutor, and the public key used for signing.
+  nix.settings.trusted-public-keys = [
+    "ghaf-jenkins:5OXpzoevBwH4sBR0S0HaIQCik2adrOrGawIXO+WADCk="
+  ];
+  nix.settings.substituters = [
+    "http://localhost:8080"
```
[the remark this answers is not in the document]
flokli: For localhost? This is only a local proxy into Azure Blob Storage, as nix doesn't support it natively.

hosts/jenkins-controller/configuration.nix
flokli: See https://github.com/tiiuae/ghaf-infra/pull/35#discussion_r1432408661

terraform/jenkins/binary_cache_signing.tf
flokli: This is a "Group" in Azure (probably an Entra Group), and I'm not sure which terraform resource can be used for that. It's probably not https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/management_group.

terraform/jenkins/binary_cache_signing.tf
flokli: You need to grant Terraform itself access to enter secrets into that vault. See the comment in L30.

terraform/jenkins/binary_cache_signing.tf
flokli: This is mostly a safeguard against terraform deleting that resource from the state, as it only lives in there, in case it gets moved around and renamed etc. It is a recommended best practice as per https://github.com/numtide/terraform-provider-secret?tab=readme-ov-file#example. 

I think the rest of this question is more one about policy, not about this code particularly.

hosts/jenkins-controller/get_secret.py
```diff
@@ -0,0 +1,24 @@
+"""
```
henrirosten, line 1: Not blocking at this point, but noting down that various checks and linters currently fail due to changes in this PR, e.g.: `reuse lint` `pylint` `inv pre-push`

henrirosten (commented):
In general, this seems good to me. Also, thanks for the overall summary in https://github.com/tiiuae/ghaf-infra/pull/35#issuecomment-1863035100 as well as providing some helpful comments throughout the changes and in the commit messages.
 
With that said, I would still like to have an 'end-user' documentation to better understand your thoughts as to what is the intended workflow: how to do a clean install, how to update hosts, how are secrets updated, how could this config be moved to other cloud providers, etc. 

What I'm a bit concerned is that we seem to pretty strictly bind ourselves to Azure. Indeed, it would be important to be able to apply, or at least understand what are the changes required to apply this configuration on other providers or bare metal installations.

Again, I realize this is still a PoC, but I still believe we should have such 'end-user' documentation available as early as possible - it would also help the review.

terraform/jenkins/builder.tf
```diff
@@ -0,0 +1,89 @@
...
+
+resource "azurerm_network_security_group" "builder_vm" {
+  count = local.num_builders
+
+  name                = "builder-vm-${count.index}"
+  resource_group_name = azurerm_resource_group.default.name
+  location            = azurerm_resource_group.default.location
+
+  security_rule {
+    name                       = "AllowSSHFromJenkins"
```
alextserepov, line 70: Is this for Jenkins controller managing builders?

terraform/jenkins/jenkins_controller.tf
```diff
@@ -0,0 +1,148 @@
...
+        ssh_authorized_keys = local.ssh_keys[user]
+      }
+    ]
+    write_files = [
+      # See corresponding EnvironmentFile= directives in services
+      {
+        content = "KEY_VAULT_NAME=${azurerm_key_vault.ssh_remote_build.name}\nSECRET_NAME=${azurerm_key_vault_secret.ssh_remote_build.name}",
+        "path"  = "/var/lib/fetch-build-ssh-key/env"
```
alextserepov, line 44: so, ssh_remote_build.name is stored as a secret in keyvault?

terraform/jenkins/jenkins_controller.tf
```diff
@@ -0,0 +1,148 @@
...
+      },
+      {
+        content = "KEY_VAULT_NAME=${azurerm_key_vault.binary_cache_signing_key.name}\nSECRET_NAME=${azurerm_key_vault_secret.binary_cache_signing_key.name}",
+        "path"  = "/var/lib/fetch-binary-cache-signing-key/env"
```
alextserepov, line 48: and again the signing key traveling somewhere ...

terraform/jenkins/jenkins_controller.tf
```diff
@@ -0,0 +1,148 @@
...
+  network_interface_id      = module.jenkins_controller_vm.virtual_machine_network_interface_id
+  network_security_group_id = azurerm_network_security_group.jenkins_controller_vm.id
+}
+
+resource "azurerm_network_security_group" "jenkins_controller_vm" {
+  name                = "jenkins-controller-vm"
+  resource_group_name = azurerm_resource_group.default.name
+  location            = azurerm_resource_group.default.location
+
+  security_rule {
```
alextserepov, line 96: why do we need inbound ssh on Jenkins controller?

terraform/jenkins/main.tf
```diff
@@ -0,0 +1,53 @@
...
+  features {}
+}
+
+terraform {
+  required_providers {
+    azurerm = {
+      source = "hashicorp/azurerm"
+    }
+    secret = {
+      source = "numtide/secret"
```
alextserepov, line 15: what is this one?

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
flokli: We don't send the key. The URL parameter is a configuration parameter for the store. It's *not* sent over the wire.

The NAR files are uploaded as well as the .narinfo files, the .narinfo file contains a signature over a *fingerprint* of some of the fields in the NAR file.

We could split the signing into a separate command if we want to.

I think long-term we should offload the signing operation to soemwhere outside the machine.

terraform/jenkins/binary_cache_signing.tf
flokli: `terraform/jenkins` is the entrypoint/root terraform module for the current deployment. An additional deployment would need to have its own terraform module (and once we start having more of these, I'd assume we'd factor out more of the common code in reusable modules).

That specific uid won't work in another azure account, but it'd come with its own vault, and where we'd instantiate it we'd decide about access policies there, and which uid should have access. It might be we don't need to give terraform access to the secret at all, if it's not terraform putting it there.

hosts/jenkins-controller/configuration.nix
```diff
@@ -0,0 +1,179 @@
...
+      Restart = "on-failure";
+    };
+    script = let
+      get-secret = pkgs.writers.writePython3 "get-secret" {
+        libraries = with pkgs.python3.pkgs; [azure-keyvault-secrets azure-identity];
+      } (builtins.readFile ./get_secret.py);
+    in ''
+      umask 077
+      mkdir -p /etc/secrets/
+      ${get-secret} > /etc/secrets/nix-signing-key
```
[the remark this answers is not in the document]
flokli: Nix is pretty bad when it comes to signing, and only supports the private key being present on some filesystem - here's the equivalent desk in the existing Hydra setup: https://github.com/tiiuae/ghaf-infra/blob/7f88348311e3dca947e4f40219968c1069e420d5/services/hydra/default.nix#L120

I agree signing shouldn't happen on the box, but instead we should offload the signing to an HSM/Key vault, so even the jenkins-controller can't see the secret. As long as that other signer can establish machine identity and supports signing arbitrary ed25519 payloads (the NARInfo fingerprint), the rest of Nix doesn't care how the signature was made.

I already wrote most of the building blocks w.r.t calculating fingerprints, adding signatures to NARInfo files etc. I'll add a futurework item for this in the docs.

terraform/jenkins/builder.tf
flokli: Yes, which is why it's restricted to the Jenkins subnet, in which only the jenkins-controller VM lives.

terraform/jenkins/jenkins_controller.tf
flokli: This configures the EnvironmentFile= that the `fetch-build-ssh-key` systemd service sees.

It sets `KEY_VAULT_NAME` to the name of the vault, and `SECRET_NAME` to the name of the secret in that vault.

Both of them are interpolated through terraform from the attributes in the corresponding resources, which properly models the dependency graph, and ensures things are created in the right order.

Rendered on disk, `/var/lib/fetch-build-ssh-key/env` currently has the following contents:

```
KEY_VAULT_NAME=ghaf-ssh-remote-build
SECRET_NAME=remote-build-ssh-private-key
```

and no, none of these values are secret, they can also be inferred by staring at the terraform code ;-)

terraform/jenkins/jenkins_controller.tf
flokli: It's not travelling, this is configuring the `fetch-binary-cache-signing-key` service, so it knows where to fetch the key from to put on disk.

Regarding "putting on disk", I refer to https://github.com/tiiuae/ghaf-infra/pull/35#discussion_r1432769524.

terraform/jenkins/jenkins_controller.tf
flokli: So I could ssh in and debug the config. Also, until Jenkins is configured declaratively and uses SSO for login and RBAC, the admin password needs to be fetched from the logs/some file in the jenkins state.

terraform/jenkins/main.tf
flokli: This is https://github.com/numtide/terraform-provider-secret, and tells terraform we mean numtide/secret, not the non-existing hashicorp/secret default.

The actual terraform provider is provided through Nix (see `devshell.nix`).

hosts/binary-cache/configuration.nix
```diff
@@ -0,0 +1,99 @@
...
+  # FUTUREWORK: rebase once https://github.com/NixOS/nixpkgs/pull/272617 landed
+  services.caddy.enableReload = false;
+  systemd.services.caddy.serviceConfig.ExecStart = lib.mkForce [
+    ""
+    "${pkgs.caddy}/bin/caddy run --environ --config ${config.services.caddy.configFile}/Caddyfile"
+  ];
+  systemd.services.caddy.serviceConfig.EnvironmentFile = "/run/caddy.env";
+
+  # Expose the HTTP and HTTPS port.
+  networking.firewall.allowedTCPPorts = [80 443];
```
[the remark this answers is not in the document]
flokli: No, this can be removed. I think I deconfigured caddy but forgot to remove the firewall entry.

hosts/jenkins-controller/get_secret.py
flokli: Yes, the in-repo linter and the linter used inside the build framework for `writers.writePython` disagree about style. I think I'll just vendor these few lines into the Nix file that makes the call.

hosts/builder/configuration.nix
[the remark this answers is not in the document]
flokli: The builders only use this to substitute build artifacts from the cache at azure blob storage (as Nix doesn't support Azure blob storage natively).

Nix does verify signatures on the NARInfo files before substituting and rejects wrong data.

Adding certificates here would make things a lot more complicated without much benefit. This connection doesn't need to be trusted. It's purely local to Nix itself, and can't be MITM'ed without breaking the signatures.

I'd prefer to keep this as-is, or if it has to be HTTPS, simply use the HTTPS interface exposed by the binary-cache VM, which does expose the same thing to external users over https, though copying that traffic in and out of the VM is a bit wasteful.

hosts/binary-cache/configuration.nix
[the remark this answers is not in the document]
flokli: Done.

hosts/builder/configuration.nix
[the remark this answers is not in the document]
alextserepov: I need to look deeper into it. Resolving for now, but let's see after the holidays.

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
alextserepov: as signing is under radar now, resolving. Will get back to it after the holidays.

hosts/jenkins-controller/configuration.nix
```diff
@@ -0,0 +1,179 @@
...
+      Restart = "on-failure";
+    };
+    script = let
+      get-secret = pkgs.writers.writePython3 "get-secret" {
+        libraries = with pkgs.python3.pkgs; [azure-keyvault-secrets azure-identity];
+      } (builtins.readFile ./get_secret.py);
+    in ''
+      umask 077
+      mkdir -p /etc/secrets/
+      ${get-secret} > /etc/secrets/remote-build-ssh-key
```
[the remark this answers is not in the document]
alextserepov: I still disagree with printing private keys to stdout. Even if stdout is redirected to a file.

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
alextserepov: As discussed, in a short term, the signing can be simply replaced with a API call to the Keyvault. The keyvault stores the key instead of secret and does the signing with the key. We send the hash over REST API for example and get the signature back. The public key can be delivered to receiving party or the same REST API call can be made to the same keyvault. The keyvault can be backed by software or by real HSM, depending purely on the budget. Naturally, access policy and firewalling must be considered here.

terraform/jenkins/binary_cache_signing.tf
alextserepov: ok.

terraform/jenkins/jenkins_controller.tf
alextserepov: needs further consideration after holidays.

terraform/jenkins/jenkins_controller.tf
alextserepov: please close on production system, unless there is other use than debug/provisioning.

terraform/jenkins/main.tf
alextserepov: Second line of readme: "Please be careful about your security stance before adopting this!" Are we careful about our security stance? ;)

terraform/jenkins/jenkins_controller.tf
flokli: Yes, there's a block in the futurework section about SSO and declarative pipeline config.
At least parts of this need to be addressed.

terraform/jenkins/main.tf
flokli: This is mostly about who has access to the state bucket, and limiting the amount of people having access to the state bucket, and feeling comfortable about having the secrets in there.

We only use it to store the signing key, if we move that into the Key Vault, delegate signing to it entirely, we might as well create the key material there without the private key material ever touching terraform, removing the need for the provider alltogether.

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
flokli: Nix doesn't support any of this yet, it insists on having the private key available as a file. That's the current state of things.

If you want to go into patching Nix, there's plans to change this, and a draft PR, but none of this is likely to land in the very short term.

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
flokli: So you'd prefer the python script having support to write to the file directly?

hosts/jenkins-controller/get_secret.py
flokli: Ci is green now, marking this as resolved.

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
alextserepov: ok, since we are not using nix-serve, as discussed, we might as well just sign in Jenkins in "post build" mode. There is nothing that would prevent us from running it as 1 extra post-build step, right?

terraform/jenkins/README.md
```diff
@@ -0,0 +1,185 @@
...
+Obviously this is very bad, as leaving the private key on the host makes it at
+least possible to steal the private key.
+
+It seems Azure Key Vault supports dealing with ed25519 key material, and doing
+signatures. Rather than keeping the key as a file on disk.
+
+A much cleaner solution would be to allow offloading the signing mechanism to
+Azure Key Vault, or "external signers in general".
+
+In terms of complexity, there's already
```
alextserepov, line 175: Even simpler solution: Generate hash & send REST API to KeyVault.

terraform/jenkins/README.md
```diff
@@ -0,0 +1,185 @@
...
+that new image.
+
+State that needs to be kept persistent for longer needs to be saved by attaching
+managed data volumes, and mounting them to the state directories of the specific
+service.
+
+### Environment-agnostic
+Images are supposed to be *environment-agnostic*, allowing multiple deployments
+to share the same image / Nix code to build it.
+
```
alextserepov, line 41: https://github.com/tiiuae/ghaf-infra/pull/35/files#r1432386111

hosts/jenkins-controller/configuration.nix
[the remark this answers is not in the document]
flokli: Signing store paths happens on all builds, and is handled by the nix daemon, not Jenkins directly.

Jenkins doesn't have access to the signing keys or any way to influence the uploading process, and this ensures all builds are captured. But we can extend/change the post-build-hook script to do something else for signing, yes.

terraform/jenkins/README.md
flokli: Nix uses a fingerprint of a NAR and requires a signature for that. There's currently no direct mechanism inside Nix to retrieve this [fingerprint[(https://docs.tvix.dev/rust/nix_compat/narinfo/fn.fingerprint.html)

flokli:
Closing, seems this all ended up in the main branch one way or another.

PR #36, opened 2023-12-08
x86 builder: Add access to binary cache and Gala app sources

mkaapu:
Add access to extra hosts:
Add access to vedenemo.dev to fetch Gala app sources.
Add access to cache.vedenemo.dev (Binary cache).

### DIFF
--- a/hosts/ficolobuild/configuration.nix
+++ b/hosts/ficolobuild/configuration.nix
@@ -36,4 +36,10 @@
   # Use the systemd-boot EFI boot loader.
   boot.loader.systemd-boot.enable = true;
   boot.loader.efi.canTouchEfiVariables = true;
+
+  # Connect hosts in the same network.
+  networking.extraHosts = "
+    172.18.20.102 vedenemo.dev # for fetching Gala app sources
+    172.18.20.109 cache.vedenemo.dev # Binary cache
+  ";
 }


hosts/ficolobuild/configuration.nix
```diff
@@ -36,4 +36,10 @@
   # Use the systemd-boot EFI boot loader.
   boot.loader.systemd-boot.enable = true;
   boot.loader.efi.canTouchEfiVariables = true;
+
+  # Connect to public servers.
```
cazfi, line 40: Contrary to the comment, this is about connecting to "private" IP instead of "public" one (which you would get from the dns)

cazfi (changes requested):

hosts/ficolobuild/configuration.nix
```diff
@@ -36,4 +36,10 @@
   # Use the systemd-boot EFI boot loader.
   boot.loader.systemd-boot.enable = true;
   boot.loader.efi.canTouchEfiVariables = true;
+
+  # Connect hosts in the same network.
+  networking.extraHosts = "
+    172.18.20.102 vedenemo.dev # for fetching Gala app sources
+    172.18.20.109 cache.vedenemo.dev # Binary cache
```
cazfi, line 43: Why you still need cache.vedenemo.dev?

cazfi (changes requested):

hosts/ficolobuild/configuration.nix
mkaapu: cache.vedenemo.dev is still in the substituters list included with hosts/common.nix to x86 builder setup. I did not start to split common.nix yet as I found out that when these builders are used as remote builders, those pre-defined substituters are not used by default. You still need to explicitly instruct remote build machines to use their own binary substitutes:
https://nixos.org/manual/nix/stable/command-ref/conf-file.html#conf-builders-use-substitutes
So, this is enabling usage of that binary cache but not forcing of its usage.

cazfi (approved):

PR #37, opened 2023-12-12
Monitor awsarm through SSH

joinemm:
[sshified](https://github.com/hoffie/sshified) can be used to proxy prometheus metrics through ssh connection, meaning we do not have to open any ports to the world.

This PR adds a derivation to build `sshified`, a systemd service to run it, and the necessary ssh configuration to scrape metrics from awsarm.

Also added documentation that tells how to add new monitored targets, with either ssh or http authentication, though there are no more targets that are using http basic auth.

### DIFF
--- a/docs/monitoring-server.md
+++ b/docs/monitoring-server.md
@@ -1,0 +1,141 @@
+<!--
+SPDX-FileCopyrightText: 2023 Technology Innovation Institute (TII)
+
+SPDX-License-Identifier: Apache-2.0
+-->
+
+# Monitoring server
+
+Monitoring server is running Grafana on port 80 and Prometheus on port 9090.
+
+It's configuration is located in `hosts/monitoring`.
+
+## Adding metrics endpoint
+
+### ubuntu host
+
+```sh
+sudo apt-get install prometheus-node-exporter
+sudo systemctl enable prometheus-node-exporter.service
+sudo systemctl start prometheus-node-exporter.service
+```
+
+Metrics should now be available at `127.0.0.1:9100/metrics`
+
+```sh
+curl 127.0.0.1:9100/metrics
+```
+
+If metrics are not accessible from another machine in ficolo network, check firewall:
+
+```sh
+sudo ufw status
+sudo ufw allow 9100/tcp
+```
+### nixos host
+
+For a host managed by this repo, simply import `service-node-exporter`
+
+## Authentication
+
+### ubuntu host
+
+For http basic auth, it is easiest to use nginx.
+
+Create password file
+
+```sh
+sudo apt-get install apache2-utils
+sudo htpasswd -c /etc/nginx/.htpasswd scraper
+```
+
+And add this to the nginx config of your server:
+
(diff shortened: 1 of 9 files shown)


hosts/binarycache/configuration.nix
```diff
@@ -56,7 +57,7 @@
     virtualHosts = {
       "cache.vedenemo.dev" = {
         enableACME = true;
-        forceSSL = true;
+        addSSL = true;
```
henrirosten, line 60: Why is this change needed again?

henrirosten (dismissed):
Looks good to me.
One item we possibly might want to simplify in a a follow-up PR:
I think we do not necessarily need the `sshified` but similar ssh tunnelling setup could be established using just `ssh` with option `-L` (or possibly `-D`). Or did you trial this already?

joinemm:
@henrirosten `ssh -L` works but the problem is we would have to keep the ssh tunnel open indefinitely, which is not very reliable. `sshified` on the other hand opens a new ssh connection for each request.

hosts/binarycache/configuration.nix
joinemm: this was to allow http connections to cache (using ip address), but maybe shouldn't be part of this pr

joinemm: removed this for now

docs/monitoring-server.md
```diff
@@ -0,0 +1,141 @@
+<!--
+SPDX-FileCopyrightText: 2023 Technology Innovation Institute (TII)
+
+SPDX-License-Identifier: Apache-2.0
```
tervis-unikie, line 4: Documentation license should be CC-BY-SA-4.0

tervis-unikie (commented):
Please make sure that the sshfied user on awsarm only has the permissions absolutely necessary for this to work.

docs/monitoring-server.md
joinemm: Good catch. Actually it looks like all of the other documentation in this repo is also using apache license which should be changed @henrirosten

joinemm:
@tervis-unikie sshified user has whatever is default permissions when using `adduser`. Should something be restricted?

docs/monitoring-server.md
henrirosten: True, we should change the documentation licenses. I propose we do this repo-wide with a separate PR after this has been merged.

henrirosten (approved):

tervis-unikie:
> @tervis-unikie sshified user has whatever is default permissions when using `adduser`. Should something be restricted?

Then the user can do whatever a normal user can do, which for the purpose of sshfied is excessive. 
This account could be set into a limited chroot environment for example.
But anyway, this is not a matter of this PR, just noting this...

docs/monitoring-server.md
tervis-unikie: I noticed this earlier this week, there's already Jira ticket about it.

PR #39, opened 2023-12-15
 Add Ficolo PRBUILDER  configuration 

karim20230:
Configure NixOS for "prbuilder" with Docker support, user management, and network settings.

### DIFF
--- a/hosts/prbuilder/configuration.nix
+++ b/hosts/prbuilder/configuration.nix
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: 2023 Technology Innovation Institute (TII)
+#
+# SPDX-License-Identifier: Apache-2.0
+{
+  self,
+  inputs,
+  lib,
+  pkgs,
+  ...
+}: {
+  imports = lib.flatten [
+    (with inputs; [
+      nix-serve-ng.nixosModules.default
+      disko.nixosModules.disko
+    ])
+    (with self.nixosModules; [
+      common
+      qemu-common
+      service-openssh
+      user-jrautiola
+      user-cazfi
+      user-karim
+    ])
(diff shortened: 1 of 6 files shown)


cazfi (approved):

hosts/prbuilder/configuration.nix
```diff
@@ -0,0 +1,39 @@
...
+      user-cazfi
+      user-karim
+    ])
+    ./disk-config.nix
+  ];
+  nixpkgs.hostPlatform = lib.mkDefault "x86_64-linux";
+   # List packages installed in system profile
+  environment.systemPackages = with pkgs; [
+    git
+    docker
```
joinemm, line 30: docker is already installed by the `virtualisation.docker.enable` option so this is not needed.

hosts/prbuilder/configuration.nix
```diff
@@ -0,0 +1,39 @@
...
+  ];
+# docker daemon running
+virtualisation.docker.enable=true;
```
joinemm, line 33: should be properly indented

joinemm (changes requested):

tasks.py
```diff
@@ -85,6 +85,7 @@ class TargetHost:
         "build3-ficolo": TargetHost(
             hostname="172.18.20.104", nixosconfig="ficolobuild"
         ),
+        "prbuilder": TargetHost(hostname="172.18.20.106", nixosconfig="prbuilder"),
```
mkaapu, line 88: minor, but maybe keep to the same style us above.

karim20230:
addressed cmt and installed  emacs to make @cazfi  life easier!

joinemm (approved):

PR #40, opened 2023-12-18
Added screen  and docker access right

karim20230:
Added screen and docker access right for Marko L

### DIFF
--- a/users/cazfi.nix
+++ b/users/cazfi.nix
@@ -8,7 +8,7 @@
       openssh.authorizedKeys.keys = [
         "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHzAww8Md+anrVfg93jNYey35Lu/YPEdbEh9QRu+riyf cazfi@cazfi-wlt"
       ];
-      extraGroups = ["wheel" "networkmanager"];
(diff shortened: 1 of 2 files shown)


users/cazfi.nix
```diff
@@ -8,7 +8,7 @@
       openssh.authorizedKeys.keys = [
         "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIHzAww8Md+anrVfg93jNYey35Lu/YPEdbEh9QRu+riyf cazfi@cazfi-wlt"
       ];
-      extraGroups = ["wheel" "networkmanager"];
+      extraGroups = ["wheel" "networkmanager" "docker"];
```
cazfi, line 11: This is now in common part of ghaf-infra, not in prbuilder specific part. Does it work correctly for other hosts, which may not even have docker group at all?

cazfi (changes requested):

users/cazfi.nix
joinemm: as far as I know this should be fine. the group is ignored if it does not exist.

cazfi (approved):

PR #41, opened 2023-12-19
Fix pre-push formatting checks

joinemm:
Current main branch does not pass `nix flake check` as the formatting is wrong in prbuilder configuration. This passed our automatic checks because the pre-push task silently reformats the files before running `nix flake check`. Thus the pre-push checks pass even though files have been changed.

`nix fmt` has no `--check` option so this was changed to ~~`alejandra --check .`~~ `nix fmt -- --fail-on-change`.

For `tasks.py`, added some commas so `black` will format the aliases consistently in the same way. Without this the wrapping would depend on the line length. Imports have also been sorted with `isort`.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -32,23 +32,22 @@
 """ Misc dev and deployment helper tasks """
 
 import json
+import logging
 import os
+import socket
 import subprocess
 import sys
-import logging
-import socket
 import time
+from collections import OrderedDict
+from dataclasses import dataclass
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Union
-from collections import OrderedDict
-from dataclasses import dataclass
 
-from tabulate import tabulate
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -491,9 +495,10 @@ def pre_push(c: Any) -> None:
     if not ret:
         LOG.warning("Run `terraform fmt -recursive` locally to fix formatting")
         sys.exit(1)
-    cmd = "nix fmt"
+    cmd = "alejandra --check ."
```
brianmcgillion, line 498: Change this to :

`nix fmt -- --fail-on-change`

brianmcgillion (changes requested):

tasks.py
joinemm: oh that's better yeah

henrirosten (approved):

PR #47, opened 2023-12-21
Add py3 dependencies for the PR builder

karim20230:
Add necessary py3 dependencies for the PR builder


mnokka (approved):
these are required pacakges

cazfi (approved):
At least step to right direction - hydractl started to work with this, so getting further in the PRBuilder setup.

joinemm (approved):

PR #50, opened 2024-01-05
Update nixpkgs to 23.11

joinemm:
23.05 has outdated grafana module which no longer works properly. Updating to 23.11 fixes the monitoring dashboard.

Requires verifying on all hosts that the update doesn't introduce breaking changes. Raising the stateVersion *should* be fine as we don't really have much state that could break.

`monitoring-ficolo` and `binarycache-ficolo` have been succesfully updated and show no recession.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1698155728,
-        "narHash": "sha256-PUJ40o/0LyMEgSBEfLVyPA0K3gQnPYQDq9dW9nCOU9M=",
+        "lastModified": 1704318910,
+        "narHash": "sha256-wOIJwAsnZhM0NlFRwYJRgO4Lldh8j9viyzwQXtrbNtM=",
         "owner": "nix-community",
         "repo": "disko",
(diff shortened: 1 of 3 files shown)


karim20230:
LGTM

karim20230 (approved):

henrirosten:
Tested on `ghafhydra-dev` and I see some services on that host fail to start after this update.

However, that server is currently not used for anything other than demonstrating that we can setup hydra and x86 builder with nixos-anywhere. Moreover, it will be replaced with https://github.com/tiiuae/ghaf-infra/pull/35 sometime in the near future, so I will not block this PR due to that failure.

If we need to, we can later inspect the issue on `ghafhydra-dev` on 23.11. Until then (until someone has time to inspect the issue in detail) `ghafhydra-dev` will keep running on 23.05.

henrirosten (approved):

mkaapu:
I think we need to test this also with Ficolo setup. We do not currently have bare metal server where we could safely test this, so I am going to deploy changes to our x86 Ficolo builder and test changes with remote building.

mkaapu (approved):

PR #57, opened 2024-01-18
Use ci-server as DNS instead of /etc/hosts entries

tervis-unikie:
Also updated copyrights of the relevant files and added ficolo DNS setup also to prbuilder.

Tested on prbuilder.

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -1,6 +1,6 @@
-# SPDX-FileCopyrightText: 2023 Technology Innovation Institute (TII)
-#
+# SPDX-FileCopyrightText: 2023-2024 Technology Innovation Institute (TII)
 # SPDX-License-Identifier: Apache-2.0
+#
 {
   self,
   inputs,
@@ -11,7 +11,7 @@
     # shared modules
     azure-common = import ./azure-common.nix;
     qemu-common = import ./qemu-common.nix;
-    ficolo-hosts = import ./ficolo-hosts.nix;
+    ficolo-common = import ./ficolo-common.nix;
     common = import ./common.nix;
     generic-disk-config = import ./generic-disk-config.nix;
   };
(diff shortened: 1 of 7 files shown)


mkaapu (approved):

karim20230 (approved):

hosts/default.nix
```diff
@@ -11,7 +11,7 @@
     # shared modules
     azure-common = import ./azure-common.nix;
     qemu-common = import ./qemu-common.nix;
-    ficolo-hosts = import ./ficolo-hosts.nix;
+    ficolo-common = import ./ficolo-common.nix;
```
cazfi, line 14: Not a new issue with this PR, but it seems wrong to me that ficolo-specific (even if common to all ficolo servers) configuration is in hosts/default.nix common to all servers.

mnokka (approved):
Logical way to change DNS

hosts/default.nix
cazfi: Thinking more about it, what used to be "wrong" turns to "breaking things" with this PR. We don't want to set those dns servers for machines outside ficolo.

tervis-unikie: Afaik this just "declares" the 'ficolo-common' as importing ficolo-common.nix.
As long as e.g. azure machines do not use 'ficolo-common' at all, they won't include any ficolo stuff into configuration.
And if this breaks things, it was broken before this PR.

cazfi (approved):

hosts/default.nix
mkaapu: Yes. This is how @tervis-unikie put it. This just 'introduces' or 'declares' a new module. You still need to include it separately to take it into use with any specific configuration.

mkaapu (approved):

PR #62, opened 2024-01-25
Ignore tmp files in main branch

joinemm:
Switching between main branch and azure-images branch leaves dirty git state because these files are only ignored on the other branch. This should fix that temporarily.


henrirosten (approved):
Thanks, LGTM

PR #63, opened 2024-01-26
Add bmg to build3

joinemm:
Adding brian to remote builder build3


karim20230 (approved):

mkaapu (approved):
I would consider Brian as a developer in build3 machine.
I would add all developer accounts to one group and list them maybe in one file, like:
`users/build3developers.nix`
and then add that group to trusted-users, that way we would include all users from that one module, and only place which we would need to edit would be that module when adding or removing new users.

But we can implement this or similar feature later.

cazfi (approved):

PR #64, opened 2024-01-29
Add themisto user to build4

mkaapu:
The build4 (x86 Ficolo build server) will be used as remote builder with Themisto Hydra.


mnokka (approved):
Go with changes

joinemm (approved):

PR #65, opened 2024-01-31
Add configuration to easily add developers to build machines

joinemm:
Adds `developers.nix` to the ficolobuild directory, where it can be imported to easily add all developers to any builder. Adding and removing developers from the file is easy, with only name and ssh keys required. The nix expression will map through this array and build the user and group configuration. 

~~I have made the decision to not create separate home directories for each developer, but instead use a shared `/home/developers` directory. The reasoning is to avoid leftover directories as developers are removed. And as this is just meant to be used as a builder, the home directory is rarely needed.~~


henrirosten (dismissed):

joinemm:
Removed the shared home directory as it caused more issues than it solved

mkaapu (approved):

karim20230 (approved):

PR #75, opened 2024-02-13
add certbot and nginx

karim20230:
Make sure so called PRBuilder VM has certbot and nginx set up.

### DIFF
--- a/hosts/prbuilder/configuration.nix
+++ b/hosts/prbuilder/configuration.nix
@@ -34,6 +34,8 @@
     git
     emacs
     screen
+    certbot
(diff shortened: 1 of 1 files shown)


cazfi (approved):

joinemm:
Certbot and nginx should be configured in the nix configuration using `services.certbot` and `services.nginx`. And we already have a base nginx configuration in `/services/nginx/` directory.

karim20230:
> services

miksi ?

joinemm:
so that everything is properly managed by nix and not manually

PR #81, opened 2024-02-23
Add arm remote builder with ubuntu image

joinemm:
- aarch64 VM added, the module is modified version of the `azurerm-linux-vm` module.
- arm machines are added as remote builders on jenkins-controller.
- Currently has lot of duplicated code, modules could be refactored in the future.


henrirosten (changes requested):
I'm getting the following error on trying to build an aarch64 target on jenkins-controller after taking the changes from this PR:
```
cannot build on 'ssh://remote-build@10.0.4.5': error: failed to start SSH connection to 'remote-build@10.0.4.5': Host key verification failed.
```
`known_hosts` on jenkins-controller is populated here: https://github.com/tiiuae/ghaf-infra/blob/bfb1c92c887d72c5ac05eddb6d6d553e791fa30a/hosts/azure/jenkins-controller/configuration.nix#L137

The builders get added to the scanlist here: https://github.com/tiiuae/ghaf-infra/blob/bfb1c92c887d72c5ac05eddb6d6d553e791fa30a/terraform/jenkins-controller.tf#L63-L64

I believe you should add the `arm_builder_vm`s to that list too.

terraform/modules/arm-builder-vm/ubuntu-builder.sh
```diff
@@ -0,0 +1,147 @@
...
+configure_builder() {
+	# Add user: nix
+	# Extra nix config for the builder,
+	# for detailed description of each of the below options see:
+	# https://nixos.org/manual/nix/stable/command-ref/conf-file
+	extra_nix_conf="
+# 20 GB (20*1024*1024*1024)
+min-free = 21474836480
+# 200 GB (200*1024*1024*1024)
+max-free = 214748364800
```
henrirosten, line 56: Please make sure this is correct in your new aarch builders too.
For instance, the comment 'Add user: nix' is clearly not correct.
Also, the arm builder disk seems to be 100GB, so the max-free of 200GB is probably not correct.

henrirosten (commented):
@joinemm: I know you did did pretty extensive trials with attempting to spin up NixOS VMs on Azure. 

Please reference the results of your trials, and point out what were the main reasons you had to go with Ubuntu+Nix instead of NixOS on aarch64 builders. 

I assume we'll later resume to this topic, and having the results from your analysis will then be valuable.

terraform/modules/arm-builder-vm/virtual_machine.tf
```diff
@@ -0,0 +1,149 @@
...
+    # azurerm_virtual_machine doesn't support the managed storage account
+    storage_uri = azurerm_storage_account.boot_diag.primary_blob_endpoint
+  }
+
+  storage_os_disk {
+    name              = "${var.virtual_machine_name}-osdisk" # needs to be unique
+    caching           = "ReadWrite"
+    create_option     = "FromImage"
+    managed_disk_type = "Standard_LRS"
+    disk_size_gb      = "100"
```
karim20230, line 62: Instead of hardcoded here, we can put the size in the variable as default!

terraform/modules/arm-builder-vm/virtual_machine.tf
```diff
@@ -0,0 +1,149 @@
...
+
+  identity {
+    type = "SystemAssigned"
+  }
+
+  os_profile {
+    computer_name = var.virtual_machine_name
+    # Unused, but required by the API. May not be root either
+    admin_username = "foo"
+    admin_password = "S00persecret"
```
karim20230, line 31: Great work Joonas! but why ps hardcoded

joinemm: copied from x86 builder so should be fine.

joinemm:
> @joinemm: I know you did did pretty extensive trials with attempting to spin up NixOS VMs on Azure.
> 
> Please reference the results of your trials, and point out what were the main reasons you had to go with Ubuntu+Nix instead of NixOS on aarch64 builders.
> 
> I assume we'll later resume to this topic, and having the results from your analysis will then be valuable.

Reasons explained now in `arm-builder-vm` readme. 

I've also updated the `ubuntu-builder.sh` script to install the same rclone service as the nixos configuration does, so it uses the azure binary cache.

henrirosten (approved):
Generally, this change looks good to me. I tested that the deployment works, and the resulting infra can build aarch64 targets. Also, the build results end-up in the deployed binary cache signed with the correct key.

However, there is one problem I noticed in testing: after taking the changes from this PR, the [deployment to UAE](https://github.com/tiiuae/ghaf-infra/tree/main/terraform#changing-azure-deploy-location) no longer works. The reason is,  the D2ps - D64ps v5 Azure image sizes are not available in UAE, so the attempt to deploy to that location fails with:

```
Error: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="The requested VM size Standard_D4ps_v5 is not available in the current region. The sizes available in the current region are: ...
```

Bigger problem is that currently, there are no Azure image sizes that support arm-based VMs on UAE.

My proposal is:
- We merge this change now despite the above issue
- We later follow-up with a work-around to address the above issue: perhaps use the aarch64 builder from another location on UAE deployments etc.

PR #84, opened 2024-02-27
Make osdisk size configurable

tervis-unikie:
Default osdisk sizes:
- jenkins-controller: 150 GB
- builder: 150 GB
- binary-cache: 50 GB

### DIFF
--- a/terraform/binary-cache.tf
+++ b/terraform/binary-cache.tf
@@ -1,5 +1,4 @@
-# SPDX-FileCopyrightText: 2023 Technology Innovation Institute (TII)
-#
+# SPDX-FileCopyrightText: 2022-2024 Technology Innovation Institute (TII)
 # SPDX-License-Identifier: Apache-2.0
 
 module "binary_cache_image" {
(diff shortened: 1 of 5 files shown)


karim20230 (approved):

mkaapu (approved):

henrirosten (approved):
@tervis-unikie : thanks for this change.

I think we should make the disk size configurable per environment, something similar to: https://github.com/tiiuae/ghaf-infra/blob/e2c07838c4d4b3110a315ecfa348570207ef42ce/terraform/main.tf#L69-L88
and a usage example, for instance in: https://github.com/tiiuae/ghaf-infra/blob/e2c07838c4d4b3110a315ecfa348570207ef42ce/terraform/jenkins-controller.tf#L26

If you prefer having this merged first, and then later follow-up with another PR to take care of making it part of environment-specific configuration options.

tervis-unikie:
That is a good idea, but I'll rather merge this and then make another PR.

PR #87, opened 2024-02-28
Perform better on low memory

henrirosten:
- Perform better when low on memory; 'priv' and 'dev' environments should be able to run the [Jenkins pipeline](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/main/ghaf-build-pipeline.groovy) by default without OOM. Also, if running out-of-memory, try to prioritize keeping the Jenkins service alive.
- Adjust image sizes so 'priv' environments should also be able to run the Jenkins pipeline by default. Currently, it seems the Ghaf target evaluation is not possible with 8GiB RAM, therefore, increasing the 'priv' Jenkins controller VM to 'Standard_E2_v5' (which has 16GiB RAM).
- Increase the binary cache VM size from 'Standard_D1_v2' to 'Standard_D2_v3' since the cost impact is relatively small (versus the cost of time spent when people need to debug possible resource issues).
- Start using memory optimized 'Standard_E' VM sizes for all jenkins-controller configurations, since nix evaluation for Ghaf is memory intensive.

### DIFF
--- a/hosts/azure/jenkins-controller/configuration.nix
+++ b/hosts/azure/jenkins-controller/configuration.nix
@@ -292,6 +292,25 @@ in {
     };
   };
 
+  # Enable early out-of-memory killing.
+  # Make nix builds more likely to be killed over more important services.
+  services.earlyoom = {
+    enable = true;
+    # earlyoom sends SIGTERM once below 5% and SIGKILL when below half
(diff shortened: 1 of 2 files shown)


ktusawrk (approved):
LGTM

PR #89, opened 2024-02-29
Fix caddy startup

henrirosten:
Move caddy environment file from `/run/caddy.env` to `/var/lib/caddy/caddy.env` to make it persist reboots. Before this change, caddy service would fail to start after the first boot due to missing caddy environment file.


karim20230 (approved):
Nice

flokli (approved):
Maybe something for the commit message. As per https://cloudinit.readthedocs.io/en/latest/reference/modules.html#write-files, Module frequency is `once-per-instance`, so `write_files` items only get written once.

TIL!

PR #91, opened 2024-03-04
Support deploying to different locations

henrirosten:
Support deploying ghaf-infra to other locations in addition to the default (North Europe).

For more details, see the documentation added in this PR: https://github.com/tiiuae/ghaf-infra/blob/b620eb8147fc72df45bd25daff4ee1e8e8189c1e/terraform/README.md#changing-azure-deploy-location


tervis-unikie (changes requested):
terraform-init.sh is not following our Bash script guidelines ( [Bash script style guide](https://github.com/tiiuae/ci-public/blob/main/docs/bash_script_style_guide.md) ). 
I know this commit is not responsible all of the terraform-init.sh but it introduces enough new code that it is reasonable to expect compliance with the guidelines.

henrirosten:
@tervis-unikie:
> terraform-init.sh is not following our Bash script guidelines ( [Bash script style guide](https://github.com/tiiuae/ci-public/blob/main/docs/bash_script_style_guide.md) ).
> I know this commit is not responsible all of the terraform-init.sh but it introduces enough new code that it is reasonable to expect compliance with the guidelines.

I didn't know we had a style guide for shell scripts. I would also argue having 31 checks for shell scripts is an overkill.

tervis-unikie:
> I didn't know we had a style guide for shell scripts. I would also argue having 31 checks for shell scripts is an overkill.

Well, we do. And changing those is something you can make a retrospective item about.

PR #97, opened 2024-03-08
Add Python and and Screen to x86 builder

mkaapu:
Adding some tools to x86 builders needed by Hydra and easy up administration work there:
- Python 3.10.13 + requests package
- GNU Screen

### DIFF
--- a/hosts/ficolobuild/builder.nix
+++ b/hosts/ficolobuild/builder.nix
@@ -45,6 +45,8 @@
   # Environment for Yubikey provisioning
   environment.systemPackages = with pkgs; [
     usbutils
+    screen
+    python310
   ];
   virtualisation.docker.enable = true;
 }
--- a/hosts/ficolobuild/builder.nix
+++ b/hosts/ficolobuild/builder.nix
@@ -46,7 +46,10 @@
   environment.systemPackages = with pkgs; [
     usbutils
     screen
-    python310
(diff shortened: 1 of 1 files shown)


hosts/ficolobuild/builder.nix
```diff
@@ -45,6 +45,11 @@
   # Environment for Yubikey provisioning
   environment.systemPackages = with pkgs; [
     usbutils
+    screen
+    (python310.withPackages (ps:
+      with ps; [
+        requests
+      ]))
```
henrirosten, line 52: Why do we need to install python 3.10 in systemPackages? What is the exact use-case here?

mnokka: What the other choices would be? Assuming that requests is needed

henrirosten: The package that needs python 3.10 and specific python libraries should state its own dependencies, rather than have them in the systemPackages.

cazfi: This is for a hydra container setup, which is not properly packaged (and won't be)

cazfi:
There's one more dependency needed: "sleep" command.

cazfi:
> There's one more dependency needed: "sleep" command.

Sorry. That's not missing from the host side. I'm facing some other problem.

cazfi (approved):

PR #98, opened 2024-03-13
nix-build.sh: correct copyright header

flokli:
This file was initially vendored from the TVL repo.

abe836b4f223ea3f3b89ce106c0d258074680589 accidentially updated the copyright header with the default TII header, but the actual copyright header needs to be preserved.


tervis-unikie (approved):
Good catch, this was not an intended change.

PR #101, opened 2024-03-19
Suggested improvements on server configurations, mainly targeting cache, but involve common settings too

tervis-unikie:
nixpkgs updated to avoid the zstd nginx plugin issue.

This is kind of a pre-review as I have not tested the changes yet.
I'd like to have a review from more epxerienced people, before I possibly do something bad on the cache (or other servers).
In any case, rollback should be possible in case of trouble.

### DIFF
--- a/hosts/binarycache/configuration.nix
+++ b/hosts/binarycache/configuration.nix
@@ -56,6 +56,7 @@
   };
 
   services.nginx = {
+    recommendedZstdSettings = true;
     virtualHosts = {
       "cache.vedenemo.dev" = {
         enableACME = true;
@@ -63,6 +64,10 @@
         default = true;
         locations."/" = {
           proxyPass = "http://${config.services.nix-serve.bindAddress}:${toString config.services.nix-serve.port}";
+          extraConfig = ''
+            zstd on;
+            zstd_types application/x-nix-archive;
+          '';
         };
       };
     };
--- a/hosts/common.nix
+++ b/hosts/common.nix
@@ -48,11 +48,24 @@ in {
       max-free = asGB 200;
       # check the free disk space every 5 seconds
       min-free-check-interval = 5;
+      # Fallback quickly if substituters are not available.
+      connect-timeout = lib.mkDefault 5;
+      # The default at 10 is rarely enough.
+      log-lines = lib.mkDefault 25;
     };
     # Garbage collection
     gc.automatic = true;
     gc.options = pkgs.lib.mkDefault "--delete-older-than 7d";
+
+    daemonCPUSchedPolicy = lib.mkDefault "batch";
+    daemonIOSchedClass = lib.mkDefault "idle";
+    daemonIOSchedPriority = lib.mkDefault 7;
   };
+
+  # Make builds to be more likely killed than important services.
+  # 100 is the default for user slices and 500 is systemd-coredumpd@
+  # We rather want a build to be killed than our precious user sessions as builds can be easily restarted.
+  systemd.services.nix-daemon.serviceConfig.OOMScoreAdjust = lib.mkDefault 250;
 
   # Sometimes it fails if a store path is still in use.
   # This should fix intermediate issues.
--- a/flake.lock
+++ b/flake.lock
@@ -141,11 +141,11 @@
     },
     "nixpkgs_2": {
       "locked": {
-        "lastModified": 1704295289,
-        "narHash": "sha256-9WZDRfpMqCYL6g/HNWVvXF0hxdaAgwgIGeLYiOhmes8=",
+        "lastModified": 1710695816,
+        "narHash": "sha256-3Eh7fhEID17pv9ZxrPwCLfqXnYP006RKzSs0JptsN84=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "b0b2c5445c64191fd8d0b31f2b1a34e45a64547d",
+        "rev": "614b4613980a522ba49f0d194531beddbb7220d3",
         "type": "github"
       },
       "original": {
--- a/services/binary-cache/default.nix
+++ b/services/binary-cache/default.nix
@@ -1,10 +1,30 @@
 # SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-{config, ...}: {
+{
+  config,
+  pkgs,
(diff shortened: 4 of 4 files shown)


hosts/binarycache/configuration.nix
```diff
@@ -54,13 +55,28 @@
...
         default = true;
         locations."/" = {
           proxyPass = "http://${config.services.nix-serve.bindAddress}:${toString config.services.nix-serve.port}";
+          extraConfig = ''
+            proxy_set_header Host $host;
+            proxy_redirect http:// https://;
+            proxy_http_version 1.1;
+            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+            proxy_set_header Upgrade $http_upgrade;
+            proxy_set_header Connection $connection_upgrade;
```
joinemm, line 75: There are options that set these:

- `services.nginx.virtualHosts.<name>.locations.<name>.recommendedProxySettings` 
  https://github.com/NixOS/nixpkgs/blob/nixos-23.11/nixos/modules/services/web-servers/nginx/default.nix#L228
  https://github.com/NixOS/nixpkgs/blob/nixos-23.11/nixos/modules/services/web-servers/nginx/default.nix#L97
  This one is actually already set globally in our nginx module.

- `services.nginx.virtualHosts.<name>.locations.<name>.proxyWebsockets`
  https://github.com/NixOS/nixpkgs/blob/nixos-23.11/nixos/modules/services/web-servers/nginx/default.nix#L439

hosts/binarycache/configuration.nix
```diff
@@ -54,13 +55,28 @@
   };
 
   services.nginx = {
+    package = pkgs.nginxStable.override {
+      modules = [pkgs.nginxModules.zstd];
+    };
+    recommendedZstdSettings = true;
```
joinemm, line 61: I think `recommendedZstdSettings` already adds the required module if the docs are to be believed: https://search.nixos.org/options?channel=23.11&show=services.nginx.recommendedZstdSettings&from=0&size=50&sort=relevance&type=packages&query=zstd

hosts/common.nix
```diff
@@ -48,12 +48,25 @@ in {
...
+
+    daemonCPUSchedPolicy = lib.mkDefault "batch";
+    daemonIOSchedClass = lib.mkDefault "idle";
+    daemonIOSchedPriority = lib.mkDefault 7;
   };
 
+  # Make builds to be more likely killed than important services.
+  # 100 is the default for user slices and 500 is systemd-coredumpd@
+  # We rather want a build to be killed than our precious user sessions as builds can be easily restarted.
+  systemd.services.nix-daemon.serviceConfig.OOMScoreAdjust = lib.mkDefault 250;
```
henrirosten, line 68: An alternative way of doing this would be something similar to what is currently done on jenkins-controller: https://github.com/tiiuae/ghaf-infra/blob/b0dbfa5e8618930d0bc9014b358187eead71ec99/hosts/azure/jenkins-controller/configuration.nix#L294-L305.

I don't have strong opinnions one way or another, as long as we can make sure this configuration actually works.

hosts/binarycache/configuration.nix
tervis-unikie: Mostly I've just adapted the settings from the source. It's really hard to see what is already set and what is not by just looking at the files in editor. Wish there was a tool helping with this. Anyway, good catch, I'll make some changes.

hosts/binarycache/configuration.nix
tervis-unikie: Again, just adapted the settings from source, not sure why they did it explicitly.

henrirosten (commented):
@flokli : if it takes you couple of minutes to say if these configuration changes make sense on a (bare-metal) binary cache, it would be good to have your say too.

hosts/binarycache/configuration.nix
flokli: See https://github.com/NixOS/nixpkgs/blob/fa9f817df522ac294016af3d40ccff82f5fd3a63/nixos/modules/services/web-servers/nginx/default.nix#L1194, you don't have to set `package`.

hosts/common.nix
flokli: Looking at https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/systemd/oomd.nix, oomd is already enabled, so slight preference over oomd than earlyoom.

tervis-unikie: So, what do you want me to do with it?

henrirosten: @tervis-unikie : keep it as it now is in ac267d9a471961b5a9bbd95cf7e5b6baa3204f79

flokli: :+1:

henrirosten (approved):

hosts/binarycache/configuration.nix
```diff
@@ -54,13 +54,18 @@
...
       "cache.vedenemo.dev" = {
         enableACME = true;
         forceSSL = true;
         default = true;
         locations."/" = {
           proxyPass = "http://${config.services.nix-serve.bindAddress}:${toString config.services.nix-serve.port}";
+          extraConfig = ''
+            zstd on;
+            zstd_types application/x-nix-archive;
+          '';
```
joinemm, line 68: This still needs `proxyWebsockets = true;` inside the location block to match the previous config.

joinemm (changes requested):

hosts/binarycache/configuration.nix
flokli: Nix doesn't know anything about websockets, we don't need to add that support.

joinemm: At least the harmonia readme example sets those config options. But it's true they might be unnecessary.

tervis-unikie: I'll leave it out for the moment.

joinemm (approved):

flokli (approved):

flokli:
I cherry-picked the harmonia fixes into here, and all PRs sent upstream and to nixpkgs (stable and unstable) got merged. We should be able to remove the override once [this is in the nixos-23.11 channel](https://nixpk.gs/pr-tracker.html?pr=297989).

This PR as it his now has been deployed for some hours, and seems to be fine. It can be merged.

henrirosten (approved):

PR #106, opened 2024-03-21
Extensible NixOS configuration

flokli:
This allows NixOS configuration to be extended with values coming purely terraform, which is then subsequently used to configure additional substituters and public keys before building an environment-specific image.


terraform/custom-nixos.nix
```diff
@@ -0,0 +1,28 @@
+# Nix trampoline calling the flake, and instantiating a
```
henrirosten, line 1: Missing license and copyright

terraform/main.tf
```diff
@@ -86,6 +86,8 @@ locals {
       osdisk_size_controller  = "150"
       num_builders_x86        = 1
       num_builders_aarch64    = 1
+      binary_cache_public_key = "foo:changeme"
```
henrirosten, line 89: We could set this to "ghaf-infra-dev:EdgcUJsErufZitluMOYmoJDMQE+HFyveI/D270Cr84I=" too for now

terraform/main.tf
```diff
@@ -97,6 +99,8 @@ locals {
       osdisk_size_controller  = "500"
       num_builders_x86        = 1
       num_builders_aarch64    = 1
+      binary_cache_public_key = "ghaf-infra-dev:EdgcUJsErufZitluMOYmoJDMQE+HFyveI/D270Cr84I="
+      binary_cache_url        = "https://changeme"
```
henrirosten, line 103: Minor detail at this point, but what do you think of setting this to 'http://localhost:8080' here, and removing the substituter setting from `hosts/azure/builder/configuration.nix` and `hosts/azure/jenkins-controller/configuration.nix`?

terraform/main.tf
```diff
@@ -97,6 +99,8 @@ locals {
       osdisk_size_controller  = "500"
       num_builders_x86        = 1
       num_builders_aarch64    = 1
+      binary_cache_public_key = "ghaf-infra-dev:EdgcUJsErufZitluMOYmoJDMQE+HFyveI/D270Cr84I="
```
henrirosten, line 102: Also not blocking anything now, but an improvement idea: support setting multiple public keys here and properly parsing such values in `terraform/modules/azurerm-nix-vm-image/nix-eval.sh`. As it is now, setting more than one value here (space separated) seems to blow-up the parsing in `terraform/modules/azurerm-nix-vm-image/nix-eval.sh`.

henrirosten (commented):
Thanks @flokli, this is great. Added a couple of small points to consider, see below.

In addition, we also need to think how to we get rid of hard-coding the substituters and public-keys [in arm builders](https://github.com/tiiuae/ghaf-infra/blob/main/terraform/modules/arm-builder-vm/ubuntu-builder.sh#L59-L60), and use the values `local.opts[local.conf].binary_cache_public_key` and `local.opts[local.conf].binary_cache_url` you are introducing with this PR on the arm builders too. We can fix the arm builders with a follow-up PR after this is merged, to not block this now.

terraform/custom-nixos.nix
flokli: WIll do

terraform/main.tf
flokli: Ok, will do

terraform/main.tf
flokli: For all machines that we define in Azure, and give (read-only) access to the bucket, I think they should definitely just use `http://localhost:8080`. There's no point in pulling that data from a neighboring machine. So for these VMs we should probably not set `extraNixSubstituter`, and just encode this in their `configuration.nix`. We define that `rclone-http` service right next to it, we know we have it there.

However, this per-environment URL should be the "public URL" (`https://` and what caddys' `SITE_ADDRESS` on the binary cache VM is pointed to). This can then be used to override `extraNixSubstituter` in places that are not in azure, or don't even have read-only access to the bucket (note `custom-nixos.nix` currently still is Azure-specific).

terraform/main.tf
flokli: Yes, parsing more structured data to Nix is painful due to the layer of bash in between, and already was very painful.

This definitely could be improved. I initially did not add it for simplicity, but we could have `nix-eval.sh` emit some more structured data (JSON), and then pass it to nix (some encoding over CLI args, or have Nix read it).

I'd like to keep this simple for now - do we really need multiple pubkeys configured on a per-environment level?

terraform/main.tf
flokli: Btw, I think some of the rclone-http setup should be living in a NixOS module, that might be enabled / configured from configuration.nix, but we shouldn't copypaste this too much into `configuration.nix`.

terraform/main.tf
flokli: Done, and added a TODO to set this to the real pubkeys used there (once we have them)

terraform/custom-nixos.nix
flokli: Done

terraform/main.tf
henrirosten: Yes, we can keep it simple now and later think if we need to support setting multiple keys etc.

flokli:
> In addition, we also need to think how to we get rid of hard-coding the substituters and public-keys [in arm builders](https://github.com/tiiuae/ghaf-infra/blob/main/terraform/modules/arm-builder-vm/ubuntu-builder.sh#L59-L60), and use the values local.opts[local.conf].binary_cache_public_key and local.opts[local.conf].binary_cache_url you are introducing with this PR on the arm builders too. We can fix the arm builders with a follow-up PR after this is merged, to not block this now.

Side note: This PR is system-aware (note how `system` itself can be explicitly set via argstr when calling `custom-nixos.nix`. So long-term we should still work on getting Arm and NixOS to work there ;-)

Regarding your request: `terraform/modules/arm-builder-vm/virtual_machine.tf` passes a script to execute to the ubuntu VM, setting everything up. We just need to tailor this to the environment.

 - rename `ubuntu-builder.sh` to `ubuntu-builder.sh.tpl`
 - add [variables](https://developer.hashicorp.com/terraform/language/expressions/strings#string-templates) for binary cache url and pubkey
 - and the terraform `templatefile()` function to render the final script, pass these values as arguments
 - introduce arguments as terraform module variables
 - call `arm-builder-vm` terraform module with these parameters from `terraform/arm-builder.tf`.

terraform/main.tf
flokli: > I'd like to keep this simple for now - do we really need multiple pubkeys configured on a per-environment level?

After thinking a bit more about this, If we have some other set of "well-known binary caches" we want to have control over enabling them or not, I'd probably rather see this being a "boolean" flag in the `custom-nixos.nix` file that enables some more options if it's set. And potentially the more detailed configuration of that living in a module, so it's better testable by CI, which executes none of this.

henrirosten (approved):

PR #107, opened 2024-03-27
Ubuntu arm-builder: Configurable trusted-public-keys

henrirosten:
Configure the trusted-public-keys in Ubuntu ARM builders based on the `binary_cache_public_key` configuration option in terraform the root module. Also, support setting the `binary_cache_url` on the Ubuntu ARM builders similarly.

For reference: this PR implements the change outlined in an earlier PR at: https://github.com/tiiuae/ghaf-infra/pull/106#issuecomment-2014565749


karim20230 (approved):
LGTM

PR #108, opened 2024-04-03
Refactor rclone-http into a module

joinemm:
Replaced copypasted rclone-http services with a reusable module, with configurable options. The caddyfile workaround is also removed as the fix has been merged into nixpkgs and backported to our version.

### DIFF
--- a/hosts/azure/binary-cache/configuration.nix
+++ b/hosts/azure/binary-cache/configuration.nix
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   self,
-  config,
   pkgs,
   lib,
   ...
@@ -10,6 +9,7 @@
   imports = [
     ../../azure-common.nix
     self.nixosModules.service-openssh
+    self.nixosModules.service-rclone-http
   ];
 
   nixpkgs.hostPlatform = lib.mkDefault "x86_64-linux";
@@ -26,38 +26,21 @@
     ];
   };
 
-  # Run a read-only HTTP webserver proxying to the "binary-cache-v1" storage
-  # container at a unix socket.
-  # This relies on IAM to grant access to the storage container.
-  systemd.services.rclone-http = {
-    after = ["network.target"];
-    requires = ["network.target"];
-    wantedBy = ["multi-user.target"];
-    serviceConfig = {
-      Type = "notify";
-      Restart = "always";
-      RestartSec = 2;
-      DynamicUser = true;
-      RuntimeDirectory = "rclone-http";
-      ExecStart =
-        "${pkgs.rclone}/bin/rclone "
-        + "serve http "
-        + "--azureblob-env-auth "
-        + "--read-only "
-        + "--addr unix://%t/rclone-http/socket "
-        + ":azureblob:binary-cache-v1";
-      # On successful startup, grant caddy write permissions to the socket.
-      ExecStartPost = "${pkgs.acl.bin}/bin/setfacl -m u:caddy:rw %t/rclone-http/socket";
-      EnvironmentFile = "/var/lib/rclone-http/env";
-    };
+  services.rclone-http = {
+    enable = true;
+    readOnly = true;
+    remote = ":azureblob:binary-cache-v1";
+    listenAddress = "unix://%t/rclone-http/socket";
   };
+
+  # On successful startup, grant caddy write permissions to the socket.
+  systemd.services.rclone-http.serviceConfig.ExecStartPost = "${pkgs.acl.bin}/bin/setfacl -m u:caddy:rw %t/rclone-http/socket";
 
   # Expose the rclone-http unix socket over a HTTPS, limiting to certain
   # keys only, disallowing listing too.
-  # TODO: use https://caddyserver.com/docs/caddyfile-tutorial#environment-variables for domain
   services.caddy = {
     enable = true;
-    configFile = pkgs.writeTextDir "Caddyfile" ''
+    configFile = pkgs.writeText "Caddyfile" ''
       # Disable the admin API, we don't want to reconfigure Caddy at runtime.
       {
         admin off
@@ -81,13 +64,6 @@
     '';
   };
 
-  # workaround for https://github.com/NixOS/nixpkgs/issues/272532
-  # FUTUREWORK: rebase once https://github.com/NixOS/nixpkgs/pull/272617 landed
-  services.caddy.enableReload = false;
-  systemd.services.caddy.serviceConfig.ExecStart = lib.mkForce [
-    ""
-    "${pkgs.caddy}/bin/caddy run --environ --config ${config.services.caddy.configFile}/Caddyfile"
-  ];
   systemd.services.caddy.serviceConfig.EnvironmentFile = "/var/lib/caddy/caddy.env";
 
   # Configure Nix to use the bucket (through rclone-http) as a substitutor.
--- a/hosts/azure/builder/configuration.nix
+++ b/hosts/azure/builder/configuration.nix
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   self,
-  pkgs,
   lib,
   ...
 }: {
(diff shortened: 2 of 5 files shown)


hosts/azure/binary-cache/configuration.nix
```diff
@@ -2,14 +2,16 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   self,
-  config,
   pkgs,
   lib,
   ...
 }: {
-  imports = [
+  imports = lib.flatten [
```
flokli, line 9: No need for flatten, just use `++` to concat the two lists, or eveb better not use the `with` at all:

```nix
[
  ../../azure-common.nix
  self.nixosModules.service-openssh
  self.nixosModules.service-rclone-http
]
```

hosts/azure/builder/configuration.nix
```diff
@@ -2,40 +2,24 @@
 # SPDX-License-Identifier: Apache-2.0
 {
   self,
-  pkgs,
   lib,
   ...
 }: {
-  imports = [
+  imports = lib.flatten [
```
flokli, line 8: same as above w.r.t. usage of flatten.

hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -73,9 +72,12 @@
       print(s.value)
     '';
 in {
-  imports = [
+  imports = lib.flatten [
```
flokli, line 75: same as above.

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+  pkgs,
+  lib,
+  config,
+  ...
+}:
+with lib; let
+  cfg = config.services.rclone-http;
+
+  addr =
+    if cfg.withCaddy
```
flokli, line 13: This introduces a slightly odd dependency / knowledge of another service that might be running too.

We can probably drop the boolean for now, in favor of a "listenAddress" module option.
On the caddy site, we can define a very precise `systemd.services.rclone-http.serviceConfig.ExecStartPost` where we allow the caddy user to access this.

And once we make use this much more frequently, we can come up with a more generic module option here (like "users allowed to access this socket" or something)

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
+# SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
```
flokli, line 1: This module should set up rclone, exposing a HTTP interface. It is too much aware of the surrounding context.

 - It should not be aware of a caddy possibly fronting it
 - it should not hardcode its remote location (:azureblob:binary-cache-v1)
 - it should not configure a Nix substituter

That way, we can reuse this in other contexts (for example, having a Jenkins using a HTTP/WebDAV interface to write artifacts)

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+      default = false;
+    };
+
+    readOnly = mkOption {
+      type = types.bool;
+      default = false;
+    };
+
+    serve = mkOption {
+      type = types.str;
```
flokli, line 31:
```suggestion
      type = types.enum ["http" "webdav"];
```

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+      default = "http";
+    };
+
+    substitutor = mkOption {
```
flokli, line 35: drop this option, this is nothing this module should be aware of

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+
+  addr =
+    if cfg.withCaddy
+    then "unix://%t/rclone-http/socket"
+    else "localhost:8080";
+in {
+  options.services.rclone-http = {
+    enable = mkEnableOption "rclone-http service";
+
+    withCaddy = mkOption {
```
flokli, line 20: drop this option

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+      type = types.bool;
+      default = false;
+    };
+
+    readOnly = mkOption {
```
flokli, line 25: add a description. Question, didn't check: we still need webdav and http if we also have this?

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+    };
+
+    substitutor = mkOption {
+      type = types.bool;
+      default = false;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    # Run a read-only HTTP webserver proxying to the "binary-cache-v1" storage
```
flokli, line 42: "proxying to a rclone remote at the confgured address"

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+        RuntimeDirectory = "rclone-http";
+        EnvironmentFile = "/var/lib/rclone-http/env";
+
+        ExecStart =
+          "${pkgs.rclone}/bin/rclone "
+          + "serve ${cfg.serve} "
+          + "--azureblob-env-auth "
+          + "${optionalString cfg.readOnly "--read-only "}"
+          + "--addr ${addr} "
+          + ":azureblob:binary-cache-v1";
```
flokli, line 62: make this an option

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+
+        # On successful startup, grant caddy write permissions to the socket.
```
flokli, line 64: Drop

services/rclone-http/default.nix
```diff
@@ -0,0 +1,75 @@
...
+        ExecStartPost = mkIf cfg.withCaddy "${pkgs.acl.bin}/bin/setfacl -m u:caddy:rw %t/rclone-http/socket";
+      };
+    };
+
+    # Configure Nix to use the bucket (through rclone-http) as a substitutor.
```
flokli, line 69: Drop

flokli (changes requested):

hosts/azure/builder/configuration.nix
```diff
@@ -46,7 +28,7 @@
   nix.settings.substituters = [
     # Configure Nix to use the bucket (through rclone-http) as a substitutor.
     # The public key is passed in externally.
-    "http://localhost:8080"
+    "localhost:8080"
```
flokli, line 31: I think nix still wants the http:// scheme.

services/rclone-http/default.nix
```diff
@@ -0,0 +1,64 @@
...
+with lib; let
+  cfg = config.services.rclone-http;
+in {
+  options.services.rclone-http = {
+    enable = mkEnableOption "rclone-http service";
+
+    listenAddress = mkOption {
+      type = types.str;
+      default = "localhost:8080";
+      description = "Which IP address and port the server should listen on";
```
flokli, line 18: unix:// is also supported, and we call it with that scheme in the binary cache config. Please extend that description to mention unix sockets are also supported. Maybe the rclone --help output has a concise description we can use here.

hosts/azure/builder/configuration.nix
joinemm: oh yes that was a typo

services/rclone-http/default.nix
joinemm: This was from the rclone docs. The --help output also has no mention of unix:// weirdly enough

flokli (approved):

hosts/azure/binary-cache/configuration.nix
```diff
@@ -26,38 +26,21 @@
...
+  # On successful startup, grant caddy write permissions to the socket.
+  systemd.services.rclone-http.serviceConfig.ExecStartPost = "${pkgs.acl.bin}/bin/setfacl -m u:caddy:rw %t/rclone-http/socket";
+
   # Expose the rclone-http unix socket over a HTTPS, limiting to certain
   # keys only, disallowing listing too.
-  # TODO: use https://caddyserver.com/docs/caddyfile-tutorial#environment-variables for domain
   services.caddy = {
     enable = true;
-    configFile = pkgs.writeTextDir "Caddyfile" ''
+    configFile = pkgs.writeText "Caddyfile" ''
```
henrirosten, line 43: Why do we change the `writeTextDir` to `writeText`?

hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -318,10 +295,15 @@ in {
...
+  services.rclone-http = {
+    enable = true;
+    protocol = "webdav";
+    remote = ":azureblob:binary-cache-v1";
+  };
+
   services.caddy = {
     enable = true;
-    configFile = pkgs.writeTextDir "Caddyfile" ''
+    configFile = pkgs.writeText "Caddyfile" ''
```
henrirosten, line 306: Why do we change the `writeTextDir` to `writeText`?

henrirosten (changes requested):
Possibly related to the below comments, in testing this configuration, I see the `caddy` service on binary-cache VM fails to start with the following error:

```
Apr 05 05:06:53 ghaf-binary-cache-henrirosten systemd[1]: Starting Caddy...
Apr 05 05:07:03 ghaf-binary-cache-henrirosten caddy[955]: {"level":"info","ts":1712293623.4792607,"msg":"using provided configuration","config_file":"/etc/caddy/caddy_config","config_adapter":"caddyfile"}
Apr 05 05:07:03 ghaf-binary-cache-henrirosten caddy[955]: Error: adapting config using caddyfile: server block without any key is global configuration, and if used, it must be first
Apr 05 05:07:03 ghaf-binary-cache-henrirosten systemd[1]: caddy.service: Main process exited, code=exited, status=1/FAILURE
Apr 05 05:07:03 ghaf-binary-cache-henrirosten systemd[1]: caddy.service: Failed with result 'exit-code'.
Apr 05 05:07:03 ghaf-binary-cache-henrirosten systemd[1]: Failed to start Caddy.
```

Did it work in your tests?

hosts/azure/jenkins-controller/configuration.nix
flokli: See the discussion in https://github.com/NixOS/nixpkgs/issues/272532.

Peeking at `nixos/tests/caddy.nix`, setting it like this should work?!

joinemm:
It seems I removed one line too many when getting rid of the workaround. `systemd.services.caddy.serviceConfig.EnvironmentFile = "/var/lib/caddy/caddy.env";` is still needed for the environment variables to load. Caddy service was failing because the `$SITE_ADDRESS` was empty.

hosts/azure/jenkins-controller/configuration.nix
joinemm: this is the new recommended way.

henrirosten (approved):

PR #109, opened 2024-04-10
hosts/*: remove lib.flatten usage

flokli:
We can simply concatenate lists with the `++` operator, no need to call lib.flatten on a list of lists to archieve the same thing.

Also, remove the surrounding with statement in case it's only used once or twice, and sort file imports before others.

### DIFF
--- a/hosts/ficolobuild/build4.nix
+++ b/hosts/ficolobuild/build4.nix
@@ -1,16 +1,13 @@
 # SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-{
-  self,
-  lib,
-  ...
-}: {
-  imports = lib.flatten [
-    (with self.nixosModules; [
+{self, ...}: {
+  imports =
+    [
+      ./builder.nix
+    ]
+    ++ (with self.nixosModules; [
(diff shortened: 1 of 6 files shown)


hosts/ficolobuild/build4.nix
```diff
@@ -1,15 +1,9 @@
 # SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-{
-  self,
-  lib,
-  ...
-}: {
-  imports = lib.flatten [
-    (with self.nixosModules; [
```
joinemm, line 9: I would maybe keep the `with` here just to be consistent with all the other hosts. Now if/when additional modules are added, this `with` block has to be refactored back in once reaching an arbitrary limit of say 3 modules

joinemm (changes requested):

hosts/ficolobuild/build4.nix
flokli: Updated.

joinemm (approved):

henrirosten (approved):

mkaapu:
Change LGTM, but how we are going to test this?

tervis-unikie (approved):

PR #112, opened 2024-04-10
Inital nixcache test

karim20230:
Inital nix cache test script

### DIFF
--- a/terraform/tests/nix_cache.sh
+++ b/terraform/tests/nix_cache.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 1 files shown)


henrirosten (changes requested):
@karim20230 : I'd propose to not work on this PR for now.
Let me suggest another way of doing something similar in the next few days, then let's decide which way we want to go.

karim20230:
Convert to draft!

henrirosten:
@karim20230:

I'd propose we use this as a baseline: https://github.com/tiiuae/ghaf-infra/pull/115.
If needed, we can add more tests to that script. Can you please review it and check that it covers the use-case you also intended?

PR #113, opened 2024-04-11
testagent: support Acroname USB Smart switches

flokli:
This adds support for the Acroname USB Smart switch connected to the system. We vendor in the expression until our nixpkgs pin moves past its inclusion into nixpkgs.

### DIFF
--- a/hosts/testagent/brainstem.nix
+++ b/hosts/testagent/brainstem.nix
@@ -1,0 +1,92 @@
+# SPDX-FileCopyrightText: 2003-2024 Eelco Dolstra and the Nixpkgs/NixOS contributors
+# SPDX-License-Identifier: MIT
+{
+  stdenv,
+  lib,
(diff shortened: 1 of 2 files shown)


hosts/testagent/brainstem.nix
```diff
@@ -0,0 +1,90 @@
+{
```
flokli, line 1: Considering it's vendored in from nixpkgs, needs an MIT SPDX Header.

henrirosten: FYI: some files in this project already use MIT license, e.g.: [here](https://github.com/tiiuae/ghaf-infra/blob/main/tasks.py#L3-L5) and [here](https://github.com/tiiuae/ghaf-infra/blob/main/shell.nix#L1-L3)

henrirosten (approved):

PR #116, opened 2024-04-12
Add Jenkins Pipeline visualization plugins

mkaapu:
The following Jenkins plugins and their dependencies are installed:
https://plugins.jenkins.io/pipeline-stage-view/
https://plugins.jenkins.io/blueocean/
https://plugins.jenkins.io/pipeline-graph-view/

### DIFF
--- a/hosts/azure/jenkins-controller/configuration.nix
+++ b/hosts/azure/jenkins-controller/configuration.nix
@@ -179,7 +179,7 @@ in {
       jenkins-auth = "-auth admin:\"$(cat /var/lib/jenkins/secrets/initialAdminPassword)\"";
     in ''
       # Install plugins
-      jenkins-cli ${jenkins-auth} install-plugin "workflow-aggregator" "github" "timestamper" -deploy
+      jenkins-cli ${jenkins-auth} install-plugin "workflow-aggregator" "github" "timestamper" "pipeline-stage-view" -deploy
 
       # Jenkins groovy config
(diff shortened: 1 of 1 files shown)


mkaapu:
We are currently investigating an optional way of configuration Jenkins using Configuration as Code:
https://plugins.jenkins.io/pipeline-stage-view/

These plugins will be added (using the current way) so we can better plan how to visualize our Pipeline project(s) with Jenkins.

If we take the Configuration as Code plugin to use then the Jenkins version, plugins and their configuration can be exported and imported using yaml files, so only plugin we would need to install using the current way would be Configuration as Code plugin and rest of the configuration would be loaded from those yaml files.

mnokka (approved):
go!

henrirosten:
@mkaapu:
Will there later be a matching [pipeline](https://github.com/tiiuae/ghaf-jenkins-pipeline) change?

mkaapu:
> @mkaapu: Will there later be a matching [pipeline](https://github.com/tiiuae/ghaf-jenkins-pipeline) change?

There is no pipeline stage view in Jenkins currently, so I decided to add these plugins so everyone can see how their pipeline stages are visualized.

There will be change which will refactor the current pipeline.

And this change will allow others also test trials like this:
https://github.com/mkaapu/ghaf-jenkins-pipeline/blob/WIP-PoC-matrix/Jenkinsfile

henrirosten (approved):

karim20230 (approved):
LGTM

PR #123, opened 2024-04-23
Refactor monitoring nix configuration

karim20230:
Refactor changes to the Prometheus scrape configurations for the ficolo-internal-node-exporter job. 

We have added static configurations for several targets, each with a unique IP address and port (9100). Each target has been assigned a label indicating the machine name

These changes will  improve the monitoring of our services by providing more detailed and specific metrics from different machines.


karim20230:
deployed and tested

joinemm (approved):

PR #124, opened 2024-04-24
az-jenkins-controller: add jenkins-upload-artifacts

flokli:
This uses the Jenkins ["HTTP POST" Plugin](https://plugins.jenkins.io/http-post/) to upload artifacts to a HTTP service, listening on localhost on the Jenkins Controller.

It internally connects to a unix domain socket, on which a `rclone` instance provides a WebDAV interface to Azure Object storage.


flokli:
The plugin ended up not doing anything at all, and after some back and forth, we might not be needing it at all. closing.

PR #125, opened 2024-04-24
Improve user configuration in test terraform script and add users 

karim20230:
Replacing the hardcoded test user configuration with a dynamic generation of user configurations
-Adding users

### DIFF
--- a/terraform/playground/test-infra.tf
+++ b/terraform/playground/test-infra.tf
@@ -117,12 +117,13 @@ module "test_vm" {
   virtual_machine_source_image = module.test_image.image_id
   virtual_machine_osdisk_size  = "50"
   virtual_machine_custom_data = join("\n", ["#cloud-config", yamlencode({
+
     users = [
-      {
-        name                = "hrosten"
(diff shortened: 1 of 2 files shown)


terraform/playground/test-infra.tf
```diff
@@ -117,13 +117,14 @@ module "test_vm" {
...
-    users = [
-      {
-        name                = "hrosten"
-        sudo                = "ALL=(ALL) NOPASSWD:ALL"
-        ssh_authorized_keys = local.ssh_keys["hrosten"]
-      },
-    ]
+   
+   users = [
+  for user in toset(["hrosten", "karim", "mkappu", "jrautiola", "mika", "cazfi"]) : {
```
mkaapu, line 122: typo: "mkappu", should be "mkaapu"

mkaapu (changes requested):

terraform/playground/test-infra.tf
karim20230: good catch.. fixed

mkaapu (dismissed):

mkaapu (approved):

mnokka (approved):
Go!

PR #128, opened 2024-04-29
Dev environment

mkaapu:
Rebasing dev-environment with main and
taking a new Jenkinsfile into use from
https://github.com/tiiuae/ghaf-jenkins-pipeline/tree/WIP-PoC-parallel


joinemm:
is there a way to avoid the duplicated commits when rebasing this?

mkaapu:
> is there a way to avoid the duplicated commits when rebasing this?
I am doing here something which was not intentional. As the real difference is only 1 line change in 1 file, I reject this PR.
Creation of duplicate commits was not intentional.

mkaapu:
Rejecting. Reason commented in PR.

PR #136, opened 2024-05-06
Allow changing terraform state file deploy location

henrirosten:
Deploy the terraform state file on the location declared when `terraform-init.sh` is executed. Before this change, the terraform state file was always deployed to North Europe, even if the rest of the infra would have been deployed to different location. Also, change `terraform-init.sh` so that it automatically switches to a private workspace - creating it if needed - to make the script usage less confusing. These changes are needed to better support the anticipated ghaf-infra deployment to other Azure locations.

Remove the playground directory and the example infra that was under the playground directory. Move the `terraform-playground.sh` to the root of the main module. It seems the dummy-test infra removed by this change did not serve its purpose for providing a simple example for how to configure nix hosts with terraform using ghaf-infra, and might have caused more confusion. Also, the test infra under `playground` was not well maintained.

Update the instructions in the `terraform/README.md` to match the above-mentioned changes.


terraform/README.md
```diff
@@ -166,28 +168,22 @@ However, ghaf-infra resources can be deployed to other Azure locations too, with
...
 
-# Deploy with Terraform apply, specifying the location with a variable:
-$ terraform apply -var="location=uaenorth" -var="envtype=dev" -auto-approve
-```
-When destroying a ghaf-infra instance deployed to a non-default location, you need to specify the location with a variable:
-```bash
-# Destroy:
-$ terraform apply -destroy -var="location=uaenorth" -auto-approve
+# Deploy with Terraform apply:
+$ terraform apply -var="envtype=dev" -auto-approve
```
karim20230, line 186: se on hyva...

terraform/README.md
```diff
@@ -151,13 +149,17 @@ $ terraform apply
...
-$ ./playground/terraform-playground.sh destroy
+# Destroy the private terraform worskapce using helper script
+$ ./terraform-playground.sh destroy
+
+# Alternatively, you can use terraform command directly
+$ terraform workspace select <workspace_name>
+$ terraform apply -destroy
 ```
-The above command removes all the resources that were created for the private development environment.
+The above command(s) remove all the resources that were created for the given environment.
```
karim20230, line 162: still isn't private?

karim20230:
My opinion - if` statements may not cover all issues. For instance if Terraform fails to initialize properly, the script will still proceed to execute the `terraform apply` command, which is likely to result in a failure.

terraform/README.md
henrirosten: Not sure I understand this comment, are you saying it's good?

terraform/README.md
henrirosten: It doesn't necessarily need to a 'private' environment, it can be any other workspace too.

terraform/README.md
karim20230: yeahI think  it's good we are adding  -auto-approve

henrirosten:
@karim20230 :
> My opinion - if statements may not cover all issues. For instance if Terraform fails to initialize properly, the script will still proceed > to execute the terraform apply command, which is likely to result in a failure.

Which change specifically are you referring there?

karim20230:
> @karim20230 :
> 
> > My opinion - if statements may not cover all issues. For instance if Terraform fails to initialize properly, the script will still proceed > to execute the terraform apply command, which is likely to result in a failure.
> 
> Which change specifically are you referring there?



> @karim20230 :
> 
> > My opinion - if statements may not cover all issues. For instance if Terraform fails to initialize properly, the script will still proceed > to execute the terraform apply command, which is likely to result in a failure.
> 
> Which change specifically are you referring there?



> @karim20230 :
> 
> > My opinion - if statements may not cover all issues. For instance if Terraform fails to initialize properly, the script will still proceed > to execute the terraform apply command, which is likely to result in a failure.
> 
> Which change specifically are you referring there?

 Yeah I've noticed that there are several `if` statements we added in the script to handle errors during the `terraform apply` (refereed top) commands  which are great! btw  I recommend adding additional error handling to catch these cases would be great.:slightly_smiling_face:

Here's an simple ex  how we could do in future! 
```
pushd "$MYDIR" >"$OUT"
if ! terraform init -backend-config="resource_group_name=$STATE_RG" -backend-config="storage_account_name=$STATE_ACCOUNT" >"$OUT"; then
    echo "[-] Failed to initialize Terraform"
    exit 1
fi
popd >"$OUT"
```

terraform/terraform-init.sh
```diff
@@ -97,10 +98,26 @@ azure_location_to_shortloc () {
...
+        STATE_RG="ghaf-infra-state-$SHORTLOC"
+        STATE_ACCOUNT="ghafinfrastate$SHORTLOC"
+    fi
+    echo -e "storage_account_rg_name=$STATE_RG\nstorage_account_name=$STATE_ACCOUNT" >"$MYDIR/.env"
     # See: ./state-storage
     pushd "$MYDIR/state-storage" >"$OUT"
     terraform init >"$OUT"
-    if ! terraform apply -auto-approve &>"$OUT"; then
+    terraform workspace select "$SHORTLOC" &>"$OUT" || terraform workspace new "$SHORTLOC" >"$OUT"
+    if ! terraform apply -var="location=$LOCATION" -auto-approve &>"$OUT"; then
```
karim20230, line 120: for instance here

henrirosten:
@karim20230:
 
> My opinion - if statements may not cover all issues. For instance if Terraform fails to initialize properly, the script will still proceed > to execute the terraform apply command, which is likely to result in a failure.

Now I think I understand what you meant. Taking this example you stated: 

```
terraform init -backend-config="resource_group_name=$STATE_RG" -backend-config="storage_account_name=$STATE_ACCOUNT" >"$OUT"
```

Notice we set [`set -e`](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#index-set) in the [beginning of the script](https://github.com/tiiuae/ghaf-infra/blob/main/terraform/terraform-init.sh#L6). That is, if `terraform init` (or any other command) fails, the script execution stops immediately, outputting the error message from the failing command.

terraform/terraform-init.sh
henrirosten: In this example, we want to continue in case the apply fails, and assume the failure is due to already existing state storage which we simply output inside the if, then continue with the script execution. If that assumption later turns out to be incorrect, the script will fail some point later.

You can see the error messages from such 'accepted' failure cases if you run the script with `-v` option.

karim20230:
that's cool!  I overlooked the `set -e`at the beginning of the script.

karim20230 (approved):

PR #140, opened 2024-05-13
Update with vnet peering configurations

fayadx-ssrc:
Added vnet peering configurations to terraform.
Included source resource group and the resource group where ghaf infra is deployed.
Source vnets and destination vnet specified with individual peering.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -177,6 +177,47 @@ resource "azurerm_subnet" "builders" {
 
 ################################################################################
 
+# Virtual network peering
+resource "azurerm_resource_group" "source-rg" {
+  name     = "RG-TII-DevOps-PROD"
+  location = "UAE North"
+}
+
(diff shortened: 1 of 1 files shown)


henrirosten:
@fayadx-ssrc : Not blocking, but please consider running `terraform fmt` on the file you changed.

Did you manage to test the configuration yet? 
I assume we should change the security rule [here](https://github.com/tiiuae/ghaf-infra/blob/ssrcdevops-deployment-test/terraform/binary-cache.tf#L84-L100) and [here](https://github.com/tiiuae/ghaf-infra/blob/04783e81221b0ec7a688704540f238c25aa12a89/terraform/jenkins-controller.tf#L111-L127) or would it be in a fllow-up PR later?

henrirosten:
After discussion with @fayadx-ssrc we concluded that this PR should be closed. New one will be pushed later if needed.

PR #142, opened 2024-05-14
Add hetzner arm builder

joinemm:
- Added configurations for Hetzner RX220 arm server. It has two 3.5TB nvme disks, of which the second one is mounted at `/nix`. 
- User accounts from awsarm are migrated into `developers.nix` which is used by both build3 and hetzarm.
- All builders have been moved under `/hosts/builders/` and files are refactored for better organization and less repeating of code. build3 and build4 should not suffer any functional changes from this refactor.


karim20230:
> Looks like our github checks do not support building aarch64 systems.
> 
> ```
> error: a 'aarch64-linux' with features {} is required to build '/nix/store/bqx0gv7hwyjc9a44wl5glnf004sadqgb-audit-disable.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, kvm, nixos-test, uid-range}
> subprocess.CalledProcessError: Command '['nixos-rebuild', 'build', '--option', 'accept-flake-config', 'true', '-v', '--flake', '.#hetzarm']' returned non-zero exit status 1.
> ```

Yeah I also noticed  btw maybe we could modify GitHub actions workflow to skip building when running on 'x86_64-linux'.

mkaapu (approved):
Changes LGTM

This change introduces a new host (arm builder) and configuration changes for 2 existing hosts. Changes need to be tested on all target hosts.

henrirosten:
@joinemm:
> Looks like our github checks do not support building aarch64 systems.
> 

True, we could make it support aarch64 targets by using a remote builder similarly to how we do it in Ghaf: https://github.com/tiiuae/ghaf/blob/f56797ad83f661bc0e9888e5926c0e891dcabbae/.github/workflows/build.yml#L168-L172.

However, that would require we authenticate the PR authors to not allow anyone to build (run their code) in our infra.
Another option would be to start using our Jenkins-based infra for the PRs pushed to this repo too.

Either way, I would recommend we add the aarch64 checks support after merging this PR, as it requires some more thought as to how do we want to do it.

For now, in this PR I would suggest removing the `hetzarm` target from the `task.py` [TARGETS table](https://github.com/tiiuae/ghaf-infra/blob/179051bc4d8270bb0086358fb6a997c47a8fb317/tasks.py#L98-L101), which would cause the `hetzarm` target to not be build as part of the [inv pre-push](https://github.com/tiiuae/ghaf-infra/blob/b7465878c9e879be78f04f77e8bc8c37d78f5dcc/.github/workflows/test-ghaf-infra.yml#L21C36-L21C48) check. We should add it back again as soon as we have a way to build aarch64 targets as part of our CI checks.

Other than that, this PR seems good to me.

cazfi (approved):

henrirosten (approved):

PR #148, opened 2024-05-16
Add hetzarm  to monitoring service

karim20230:
-Adding hetzarm  to monitoring service  to ensure that our monitoring service is tracking the performance and status of hetzarm. 
-Currently, our monitoring service also tracking awsarm, but this is no longer necessary. so removing awsarm from our monitoring tools to focusing our resources on active and relevant services.


karim20230:
-We agreed to stop monitoring AWSARM only when we are really ready to decommission it. so closing it and 
now Adding hetzarm to monitoring with Sshified which provides secure access and monitoring of the Hetzarm server.
[new pr](https://github.com/tiiuae/ghaf-infra/pull/149)

PR #152, opened 2024-05-17
hosts/testagent: downgrade brainstem to 2.9.25

flokli:
It seems 2.10.5 is unable to change the upstream port:

```
AcronameHubCLI -u 1
Sucessfully connected to USBHub3p SN: 0xEE92E4FD
There was an error (2) changing the upstream port
```

Revert to 2.9.25, which doesn't expose this issue.

Unfortunately, upstream doesn't seem to host that version on their website anymore (?), or at least the URL looks different, so use the one we vendored into the `ci-test-automation` repo for now.

### DIFF
--- a/hosts/testagent/brainstem.nix
+++ b/hosts/testagent/brainstem.nix
@@ -4,7 +4,7 @@
   stdenv,
   lib,
   autoPatchelfHook,
-  fetchzip,
+  fetchFromGitHub,
   curl,
   systemd,
   zlib,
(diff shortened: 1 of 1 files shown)


flokli:
I manually built and tested `/nix/store/4w51l28zv2j5w4a7d1bz8a6cd3318lnb-brainstem-2.9.25/bin/AcronameHubCLI`, and it is able to change the upstream port.

henrirosten (approved):

PR #155, opened 2024-05-21
Adds jenkins connection and mariia user

vjuntunen:
Adds connection to controller as a systemd service. Adds mariia user to ghaf-infra configuration
and testagent host.

### DIFF
--- a/hosts/testagent/configuration.nix
+++ b/hosts/testagent/configuration.nix
@@ -29,6 +29,7 @@ in {
       user-vjuntunen
       user-flokli
       user-jrautiola
+      user-mariia
     ]);
 
   # Bootloader.
@@ -48,6 +49,7 @@ in {
   environment.systemPackages = [
     inputs.robot-framework.packages.${pkgs.system}.ghaf-robot
     brainstem
+    pkgs.minicom
   ];
 
   # Disable suspend and hibernate - systemd settings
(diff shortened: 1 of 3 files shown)


joinemm:
You need to fix the formatting with `nix fmt` to pass the checks

hosts/testagent/configuration.nix
```diff
@@ -75,6 +77,36 @@ in {
...
+    pkgs.jq
+    pkgs.csvkit
+    pkgs.sudo
+    brainstem
+    inputs.robot-framework.packages.${pkgs.system}.ghaf-robot
+    ];
+    serviceConfig = {
+      Type = "simple";
+      User = "vjuntunen";
+      WorkingDirectory = "/home/vjuntunen/Jenkins-agent/";
```
henrirosten, line 104: Is this really the user and working directory we want to run this? Why?

hosts/testagent/configuration.nix
```diff
@@ -75,6 +77,36 @@ in {
...
+      ExecStart = ''${pkgs.jdk}/bin/java -jar agent.jar -jnlpUrl https://ghaf-jenkins-controller-villepekkajuntun.northeurope.cloudapp.azure.com/computer/testagent/jenkins-agent.jnlp -secret @secret-file -workDir "/home/vjuntunen/Jenkins-agent"'';
```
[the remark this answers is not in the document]
vjuntunen: If you remember when I was demoing this and we agreed that this will be manual step. So it needed to refer to env. that is wanted to be connected.

hosts/testagent/configuration.nix
vjuntunen: Well for now it is and I have different task to change it to Jenkins user and maybe better to do that with own PR?. Like we have talked in our meetings.

mkaapu (approved):

PR #156, opened 2024-05-22
Add aarch64 emulation for builders using binfmt

joinemm:
- This makes it possible to build `#packages.aarch64-linux.nxp-imx8mp-evk-debug` on x86
- Also simplified trusted users configuration.

### DIFF
--- a/hosts/builders/build3/configuration.nix
+++ b/hosts/builders/build3/configuration.nix
@@ -12,6 +12,7 @@
   imports =
     [
       ../ficolo.nix
+      ../cross-compilation.nix
       ../developers.nix
       ../yubikey.nix
       inputs.sops-nix.nixosModules.sops
@@ -43,8 +44,6 @@
 
   nix = {
     settings = {
-      # add ability to build 32 bit
-      extra-platforms = ["i686-linux"];
       trusted-users = ["@wheel" "build3"];
     };
 
--- a/hosts/builders/build3/configuration.nix
+++ b/hosts/builders/build3/configuration.nix
@@ -43,11 +43,9 @@
 
   nix = {
     settings = {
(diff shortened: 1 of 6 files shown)


hosts/common.nix
```diff
@@ -20,6 +20,9 @@ in {
     nixPath = lib.mapAttrsToList (key: value: "${key}=${value.to.path}") config.nix.registry;
 
     settings = {
+      # root is trusted by default, but only when other users aren't specufied.
```
cazfi, line 23: Typo in "specified"

cazfi (approved):

hosts/common.nix
joinemm: fixed

henrirosten:
Don't we need this for the azure builders too?

hosts/builders/build3/configuration.nix
```diff
@@ -43,11 +44,7 @@
 
   nix = {
     settings = {
-      # avoid copying stuff over ssh
-      builders-use-substitutes = true;
```
mkaapu, line 47: Not sure about this. This builder is intended to be used mostly as a remote builder. This is useful when build3 has faster internet connection than your local machine. Not sure if this is the case mostly.

mkaapu (approved):

hosts/builders/build3/configuration.nix
joinemm: this setting is already set to true in common.nix so it's not required to be here

joinemm:
@henrirosten 
> Don't we need this for the azure builders too?

yes that is coming later in another pr

karim20230 (approved):

henrirosten (approved):

PR #158, opened 2024-05-23
Use new Jenkinsfile to visualize Nightly Builds

mkaapu:
Taking Jenkinsfile from WIP-PoC-parallel branch into use with nightly builds where parallel stage directives are used to visualize target testing.


hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -159,11 +159,10 @@ in {
                   git = {
                     url = "https://github.com/tiiuae/ghaf-jenkins-pipeline.git";
                     clean = true;
-                    branches = ["*/main"];
+                    branches = ["*/WIP-PoC-parallel"];
```
henrirosten, line 162: I would suggest to merge the changes from https://github.com/tiiuae/ghaf-jenkins-pipeline/tree/WIP-PoC-parallel to the [main](https://github.com/tiiuae/ghaf-jenkins-pipeline/tree/main) branch pipeline [ghaf-nightly-build.groovy](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/main/ghaf-nightly-build.groovy). Meaning, this PR would not be needed at all.

Otherwise, if we start pulling different pipelines from different ghaf-jenkins-pipeline branches, it soon gets really hard to keep track of which pipeline version we are really using.

What's the reason we would not merge the pipeline changes to main?

mkaapu: > What's the reason we would not merge the pipeline changes to main?
The current pipeline in main branch provides different approach and more simple structure for Jenkinsfile.
It might be beneficial to keep these different solution in different branches until we have chosen the right approach to continue with.
We could also add this Jenkinsfile (with different name, maybe) to main, but still keep current nightly build version there until we choose with which version we really want to take into production.
Branches are just kind of natural way to experience with different approaches and variants.

henrirosten: > Branches are just kind of natural way to experience with different approaches and variants.

I agree, but in the ghaf-infra main branch, I would prefer pulling from the ghaf-jenkins-pipeline main to keep it simple.
You can always PoC in your private ghaf-infra deployment pulling from different branches.

> We could also add this Jenkinsfile (with different name, maybe) to main, but still keep current nightly build version there until we choose with which version we really want to take into production.

If you need to keep both, then my preference would be what you suggested above.

mkaapu: > If you need to keep both, then my preference would be what you suggested above.
At this point, I want to keep all 3 approaches (simple, parallel, matrix) so we can continue PoC development of each.

I will add this Jenkinsfile to the ghaf-jenkins-pipeline main as a separate pipeline script for nightly builds with a new name, without removing the current one.

mnokka-unikie (approved):
go

PR #160, opened 2024-05-24
jenkins: migrate to configuration-as-code, enable html markup

flokli:
The configuration-as-code plugin is the recommended way to configure
    Jenkins, not the Groovy script. Most plugin settings seem to be
    available through that.
    
    The yaml was produced by checking in the dump created by the plugin.

---

    azure/jenkins-controller: enable safe-html
    
    This allows configuring HTML descriptions for build descriptions, which
    we'll use to link to artifacts in the future.

PR #161, opened 2024-05-24
jenkins: preparations to upload and serve artifacts

flokli:
Depends on #160, will be rebased once merged (so ignore the first 2 commits).

This adds the necessary services running on the jenkins-controller machine to upload and serve artifacts from Azure storage.

 - A per-environment storage container for artifacts is added via terraform
 - Two rclone services exposing the bucket contents, rw and ro, plus config to expose the ro part via caddy.
 - Our rclone gains another patch/feature, support to connect to webdav backends via unix domain sockets, and that rclone is made available in jenkins' `$PATH`.

### DIFF
--- a/hosts/azure/jenkins-controller/configuration.nix
+++ b/hosts/azure/jenkins-controller/configuration.nix
@@ -327,6 +327,8 @@ in {
         "${rclone}/bin/rclone"
         "serve"
         "webdav"
+        "--dir-cache-time"
+        "5s"
         "--azureblob-env-auth"
         "--disable-dir-list"
         ":azureblob:jenkins-artifacts-v1"
@@ -359,6 +361,8 @@ in {
         "serve"
         "http"
         "--read-only"
+        "--dir-cache-time"
+        "5s"
         "--azureblob-env-auth"
         ":azureblob:jenkins-artifacts-v1"
       ];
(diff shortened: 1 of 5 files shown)


henrirosten:
@flokli: On testing this change on a new deployment, the Jenkins service seems to boot-up to the setup wizard. Is that expected?

flokli:
Hmm, I only tested with my installation, keeping its state. We might need to set `-Djenkins.install.runSetupWizard=false` as per [here](https://github.com/jenkinsci/configuration-as-code-plugin/issues/393#issuecomment-408122043).

flokli:
Disabling the setup wizard is done in https://github.com/tiiuae/ghaf-infra/pull/160, this is unrelated to this PR.

flokli:
Drafting until #160 is merged.

flokli:
With https://github.com/tiiuae/ghaf-infra/pull/160 merged, I rebased this.

flokli:
Corresponding pipeline PR: https://github.com/tiiuae/ghaf-jenkins-pipeline/pull/25

henrirosten:
Reviewed and verified working by deploying to a private environment.

I guess once this is merged, and all the Jenkins pipelines are changed to no longer use the `archiveArtifacts`, we could also downsize the [Jenkins disk](https://github.com/tiiuae/ghaf-infra/blob/e0fe744217d49ba643207b7fe982e872907157ba/terraform/jenkins-controller.tf#L136) to something more reasnoable. But that needs to happen in a follow-up PR later.

henrirosten (approved):

flokli:
Yeah, that's blocked on everything being switched over, and we might need to schedule some time to properly sync some existing state (if we want to keep the disk contents)

PR #172, opened 2024-06-05
Add workflow version updater

karim20230:
Adding GitHub Actions workflow to automatically update the versions of actions used in the ghaf-infra. The workflow runs on a weekly schedule and uses the 'saadmk11/github-actions-version-updater' action.

FYI: A pull request will be created with the title "ci: Update GitHub Actions to Latest Version" if updates are available.

### DIFF
--- a/.github/workflows/schedule-update-actions.yml
+++ b/.github/workflows/schedule-update-actions.yml
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+
+name: GitHub Actions Version Updater
+on:
+  workflow_dispatch:
+  schedule:
+    # Automatically run on every Sunday at 00:00 UTC.
+    - cron:  '0 0 * * 0'
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3.5.2
+        with:
+          # [Required] Access token with `workflow` scope.
+          token: ${{ secrets.PAT }}
+      
+      - name: Run GitHub Actions Version Updater
+        uses: saadmk11/github-actions-version-updater@v0.8.1
+        with:
+          # [Required] Access token with `workflow` scope.
+          token: ${{ secrets.PAT }}
+          pull_request_title: "Update GitHub Actions to Latest Version"


karim20230:
PAT needs to be added in the Repo with `workflow` scope.

mkaapu (approved):

.github/workflows/schedule-update-actions.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+
+    steps:
+      - uses: actions/checkout@v3.5.2
+        with:
+          # [Required] Access token with `workflow` scope.
+          token: ${{ secrets.PAT }}
```
henrirosten, line 19: I don't think actions/checkout should need any tokens?
It's already used e.g. here: https://github.com/tiiuae/ghaf-infra/blob/main/.github/workflows/test-ghaf-infra.yml#L18 with no token.

.github/workflows/schedule-update-actions.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - name: Run GitHub Actions Version Updater
+        uses: saadmk11/github-actions-version-updater@v0.8.1
+        with:
+          # [Required] Access token with `workflow` scope.
+          token: ${{ secrets.PAT }}
+          pull_request_title: "ci: Update GitHub Actions to Latest Version"
```
henrirosten, line 26: If we would enable `skip_pull_request` to not push PRs for updates, would it still also require the token?

I'd suggest using `update_version_with: 'release-commit-sha'` as stated in https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions and also https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

henrirosten (dismissed):
Did you test this in a fork somewhere?
If not, please do, and provide a link where we could check how it works.

karim20230:
> Did you test this in a fork somewhere? If not, please do, and provide a link where we could check how it works.

sure, here's you can take a look the eg
https://github.com/karim20230/test/pull/5
also
https://github.com/Katulos/vcf2ldif/pull/39

.github/workflows/schedule-update-actions.yml
karim20230: Great point, I'll test this btw based on their docs  Access token was required..

.github/workflows/schedule-update-actions.yml
karim20230: Yes!! would still likely require a token  .. :bowtie:

karim20230:
FYI: I have received ssrcdevops PAT from  the DEVSecOpsUAE team! The PAT has been assigned workflow permissions and has access on the  target repository, Therefore, we don't need to add a user. I've already added the PAT, so we're ready to proceed.

joinemm:
Do you have a link to example PR what this workflow would create?

karim20230:
> Do you have a link to example PR what this workflow would create?

sorry I didn't notice your comment,  sure here's you can see the example PR https://github.com/karim20230/ghaf-infra/pull/9/files

joinemm (approved):

PR #171, opened 2024-06-05
Add temporary remote-build user

henrirosten:
Add a temporary user for Azure dev remote builds on hetzarm (hetzarm.vedenemo.dev) and ficolo (builder.vedenemo.dev).

PR #173, opened 2024-06-05
terraform/jenkins-controller: add static builders

flokli:
This adds builder.vedenemo.dev and hetzarm.vedenemo.dev as static builders to /etc/nix/machines, as well as the keyscan machinery.

https://github.com/tiiuae/ghaf-infra/pull/171 already ensured that ssh key works.

maxJobs is set to the number of cores the system sees, and speedFactor is set to 3, making it more attractive than the other builders.


flokli:
Updated, it's now conditionalized on `(var.envtype == "dev" || var.envtype == "priv")`. PTAL.

joinemm (approved):

henrirosten:
We might actually want to set the [num_builders](https://github.com/tiiuae/ghaf-infra/blob/d8d62befe29da13cf4c855f7f8facf0cd27b0555/terraform/main.tf#L106-L107) to 0 on 'priv' and 'dev' configurations either as part of this PR, or on a follow-up PR.

flokli:
Let's do it in a followup, it's unrelated to adding static builders (what this PR does)

PR #180, opened 2024-06-11
Add aarch64 checks and auth to workflow

joinemm:
- Added hetzarm as remote builder so aarch64 configs can be built.
- Added build3 as remote builder to speed up building of the x86 targets.
- Added authentication steps to the workflow now that we are dealing with secrets and have to use `pull_request_target`. The auth steps are similar what is being used in ghaf repo but simplified. Reasoning explained here https://ssrc.atlassian.net/wiki/x/OABjMg

FYI the checks for this PR will fail as it's not able to use the secrets yet.

Tests done in my fork using alt account to make different PRs https://github.com/joinemm/ghaf-infra/pulls

### DIFF
--- a/.github/workflows/test-ghaf-infra.yml
+++ b/.github/workflows/test-ghaf-infra.yml
@@ -1,23 +1,62 @@
 # SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
 
-name: Test ghaf-infra
+name: Run pre-push checks
 
 on:
   push:
     branches:
       - main
-  pull_request:
+  pull_request_target:
     branches:
       - main
 
 jobs:
-  tests:
+  # Checks if the author of pull request is in our predefined list of authorized users
+  check-identity:
+    runs-on: ubuntu-latest
+    outputs:
+      authorized_user: ${{ steps.check-authorized-user.outputs.authorized_user}}
+    environment: 'internal'
+    steps:
+      - name: Check identity
+        id: check-authorized-user
+        shell: bash
+        run: |
+          # AUTHORIZED_USERS is a newline separated list of usernames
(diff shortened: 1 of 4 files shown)


.github/workflows/test-ghaf-infra.yml
```diff
@@ -1,21 +1,110 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request_target' }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
+          fetch-depth: 0
+
+      - name: Check if workflow is modified
+        id: build-yml-changed
```
henrirosten, line 62: IMHO, the name `build-yml-changed` is confusing here. Please consider renaming to something like `this-yml-changed` or `self-yml-changed` or `test-ghaf-infra-yml-changed` etc.

.github/workflows/test-ghaf-infra.yml
```diff
@@ -1,21 +1,110 @@
...
+        uses: tj-actions/changed-files@v40
+        with:
+          files: .github/workflows/test-ghaf-infra.yml
+
+      - name: Send warning
+        run: |
+          if [ "${{ steps.build-yml-changed.outputs.any_changed }}" == "true" ]; then
+            echo "::error::"\
+                 "This change edits workflow file '.github/workflows/test-ghaf-infra.yml'."\
+                 "Raising this error to notify that the build workflow change will only take impact after merge."\
```
henrirosten, line 72: Same thing here too, consider removing the reference to 'build' workflow in the comment, for example to:

"Raising this error to notify that the workflow change will only take impact after merge."

henrirosten (approved):
In general, this seems fine but I don't have means to test this right now.
I assume @joinemm that you have been fully testing this in your fork somewhere. If that's the case, I'm fine with merging this.

PR #187, opened 2024-06-20
Fix whitespace issues in workflow

joinemm:
Username was not being detected properly because the indentation of the script was causing additional whitespace on the lines when echoed. This whitespace caused grep to not match. The fix is replace any repeated whitespace with newlines.


cazfi (approved):

henrirosten (approved):
LGTM, again assuming this has been properly tested elsewhere.

PR #190, opened 2024-06-26
Reduce jenkins controller state disk size to 64gb

joinemm:


flokli (approved):
This might need backing up the other state on these boxes re-creation and restore of the volume contents (at least on staging/prod), doubt Azure can do a live-shrink of the underlying filesystem.

mnokka (approved):
go

cazfi (dismissed):

henrirosten (changes requested):
@joinemm : We *do not* want to change the osdisk size, but the jenkins controller state disk size can now be reduced to something more reasonable. The correct property we want to shrink is this one: https://github.com/tiiuae/ghaf-infra/blob/dd7509aeaea1ffcaa090846bbdfdec5471bfe4ef/terraform/jenkins-controller.tf#L142

joinemm:
@henrirosten oh ok, good catch. Fixed.

henrirosten (approved):

PR #191, opened 2024-06-26
Use 64 core builders in prod environment

joinemm:
Prod is only created for the duration of the build so the cost increase is negligible. This should speed up the builds.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -139,7 +139,7 @@ locals {
       vm_size_builder_x86     = "Standard_D64_v3"
       vm_size_builder_aarch64 = "Standard_D64ps_v5"
       osdisk_size_builder     = "500"
-      vm_size_controller      = "Standard_E4_v5"
+      vm_size_controller      = "Standard_E16_v5"
       osdisk_size_controller  = "1000"
       num_builders_x86        = 1
       num_builders_aarch64    = 1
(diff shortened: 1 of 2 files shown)


cazfi (dismissed):

mnokka (approved):
go

terraform/main.tf
```diff
@@ -136,8 +136,8 @@ locals {
     prod = {
       vm_size_binarycache     = "Standard_D2_v3"
       osdisk_size_binarycache = "250"
-      vm_size_builder_x86     = "Standard_D16_v3"
-      vm_size_builder_aarch64 = "Standard_D16ps_v5"
+      vm_size_builder_x86     = "Standard_D64_v3"
+      vm_size_builder_aarch64 = "Standard_D64ps_v5"
       osdisk_size_builder     = "500"
       vm_size_controller      = "Standard_E4_v5"
```
henrirosten, line 142: While we are changing this, let's increase the jenkins-controller size also on the prod. Maybe to "Standard_E8_v5" or even "Standard_E16_v5". Nix evaluation happens on the controller, so I assume growing its size would also help.

joinemm: Increased to 16 core

henrirosten (approved):

PR #204, opened 2024-07-22
Add a ssh-key for tervis@tervis-servu

tervis-unikie:
Allow tervis to access machines also from his desktop machine.


mkaapu (approved):
Tervis is still user, aka administrator in ghaf-infra.
We can live with that now. And if we so decide we can move Tervis to
https://github.com/tiiuae/ghaf-infra/blob/main/hosts/builders/developers.nix
instead later.
Approving.

joinemm (approved):

PR #208, opened 2024-07-31
Move tervis from admin users to developers

joinemm:
Removed tervis admin user and added to developers.nix.

Someone with decryption keys for `testagent/secrets.yaml` (@vjuntunen or @flokli) has to run `sops updatekeys` on it after these changes, I apparently do not have (also added myself in this PR)


mkaapu (approved):
LGTM

PR #215, opened 2024-08-07
Use deploy-rs

joinemm:
Adds deploy-rs deployment tool, and the necessary configuration in `nix/deployments.nix`

### DIFF
--- a/nix/deployments.nix
+++ b/nix/deployments.nix
@@ -1,0 +1,40 @@
+# SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  self,
+  inputs,
+  ...
+}: let
+  inherit (inputs) deploy-rs;
+
+  mkDeployment = arch: config: hostname: {
+    inherit hostname;
+    profiles.system = {
+      user = "root";
+      path = deploy-rs.lib.${arch}.activate.nixos self.nixosConfigurations.${config};
+    };
+  };
+
+  x86-nodes = {
+    build3 = mkDeployment "x86_64-linux" "build3" "172.18.20.104";
+    build4 = mkDeployment "x86_64-linux" "build4" "172.18.20.105";
+    himalia = mkDeployment "x86_64-linux" "himalia" "172.18.20.106";
+    monitoring = mkDeployment "x86_64-linux" "monitoring" "172.18.20.108";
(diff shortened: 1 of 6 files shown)


henrirosten:
LGTM, assuming this has been properly tested.

henrirosten (approved):

nix/deployments.nix
```diff
@@ -0,0 +1,42 @@
...
+{
+  self,
+  inputs,
+  ...
+}: let
+  inherit (inputs) deploy-rs;
+
+  mkDeployment = arch: config: hostname: {
+    inherit hostname;
+    remoteBuild = arch != "x86_64-linux";
```
flokli, line 12: This seems unnecessary, and could cause a lot of load to non-x86 machines.

If the closure to deploy is either entirely uploaded to the binary cache, or the machine doing the deployment has a aarch64 builder configured, it should be possible to drop it.

I'd prefer to not set this, and document that there exists a `--remote-build` CLI flag in deploy-rs in cases where you don't have an aarch64-linux builder configured on the machine you deploy from, and it's not in the binary cache.

nix/deployments.nix
```diff
@@ -0,0 +1,42 @@
...
+  };
+
+  aarch64-nodes = {
+    hetzarm = mkDeployment "aarch64-linux" "hetzarm" "65.21.20.242";
+  };
+in {
+  flake = {
+    deploy.nodes = x86-nodes // aarch64-nodes;
+
+    # We cannot run checks for the aarch64 nodes on an x86 system
```
flokli, line 36: We can (if there's a remote builder setup), though this comment doesn't directly relate to anything below. We define both checks. I'd propose just removing that comment to prevent confusion.

flokli (approved):
Apart from some nits around aarch64, looks nice to me!

I'd probably even go further and propose removing the other deployment commands from `invoke` and `tasks.py` - but that can probably be in a followup, once everyone gained some confidence with the tool.

nix/deployments.nix
joinemm: You're right this comment doesn't make a lot of sense without the context of what the default deploy-rs checks looked like. Removing.

nix/deployments.nix
joinemm: I wanted everything to work by default on x86 without configured remote builders, but in the scope of this project maybe it's fair to assume those exist.

flokli (approved):
Awesome, thanks!

PR #216, opened 2024-08-08
main.tf: Add hard-coded name for dev bincache

henrirosten:
This change adds another name for the 'dev' binary cache: `dev-cache.vedenemo.dev`.

The value of environment-specific terraform variable 'binary_cache_url' is read in the Caddy configuration from the SITE_ADDRESS variable, populated on the binary cache host via the `/var/lib/caddy/caddy.env` file.

Caddy supports comma separated site addresses as documented in: https://caddyserver.com/docs/caddyfile/concepts#addresses so we can directly use the same syntax in the 'binary_cache_url' terraform variable.

Caddy will also automatically take care of obtaining and renewing Let's Encrypt certificates for all the site addresses: https://caddyserver.com/docs/automatic-https.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -126,7 +126,12 @@ locals {
       num_builders_aarch64    = 0
       # 'priv' and 'dev' environments use the same binary cache signing key
       binary_cache_public_key = "ghaf-infra-dev:EdgcUJsErufZitluMOYmoJDMQE+HFyveI/D270Cr84I="
-      binary_cache_url        = "https://ghaf-binary-cache-${local.ws}.${azurerm_resource_group.infra.location}.cloudapp.azure.com"
+      # TODO: adding multiple urls as comma-and-whitespace separated
+      # string is more or less a hack. If we plan to have multiple domains
+      # per host permanently, we could make the below variable a list, and
+      # do the templating to a comma-and-whitespace separated list of urls
+      # before we pass it to caddy.
+      binary_cache_url = "https://ghaf-binary-cache-${local.ws}.${azurerm_resource_group.infra.location}.cloudapp.azure.com, https://dev-cache.vedenemo.dev"
       ext_builder_machines = [
         "ssh://remote-build@builder.vedenemo.dev x86_64-linux /etc/secrets/remote-build-ssh-key 32 3 kvm,nixos-test,benchmark,big-parallel - -",
         "ssh://remote-build@hetzarm.vedenemo.dev aarch64-linux /etc/secrets/remote-build-ssh-key 40 3 kvm,nixos-test,benchmark,big-parallel - -"


henrirosten:
This change is now deployed to dev binary cache:
```
❯ curl -L https://dev-cache.vedenemo.dev/nix-cache-info
StoreDir: /nix/store
````

or:

```
❯ curl -Ls https://dev-cache.vedenemo.dev/qyrz2xdkr921mwkyb1zfq19gi7dybbb4.narinfo

StorePath: /nix/store/qyrz2xdkr921mwkyb1zfq19gi7dybbb4-example
URL: nar/0080rx41d8lvhdqid5vabpa3wmpamrzy0qp6wq6q2l8g8jwkkcqs.nar.zst
Compression: zstd
FileHash: sha256:0080rx41d8lvhdqid5vabpa3wmpamrzy0qp6wq6q2l8g8jwkkcqs
FileSize: 95
NarHash: sha256:0xx3hgianj0d99gx0fr0ph8mall66rdiybgngsacx9hsk9nzbmw8
NarSize: 128
References: 
Deriver: iy32lrn2fhf3rx4d5pfvh0ivhfcfn6fz-example.drv
Sig: ghaf-infra-dev:Mp8QaBkDSjx3T8QAO7YzGYqouKUlror5IGA1hWu5DeqoTODe4SYABtb5VDXPO+9oVI8ZZoDvCorX269xhDE9Ag==
```

File: terraform/main.tf [hunk 1]
```diff
@@ -126,7 +126,7 @@ locals {
       num_builders_aarch64    = 0
       # 'priv' and 'dev' environments use the same binary cache signing key
       binary_cache_public_key = "ghaf-infra-dev:EdgcUJsErufZitluMOYmoJDMQE+HFyveI/D270Cr84I="
-      binary_cache_url        = "https://ghaf-binary-cache-${local.ws}.${azurerm_resource_group.infra.location}.cloudapp.azure.com"
+      binary_cache_url        = "https://ghaf-binary-cache-${local.ws}.${azurerm_resource_group.infra.location}.cloudapp.azure.com, https://dev-cache.vedenemo.dev"
```
flokli, line 129: This is a bit of a hack and not an URL - this only works "by accident" as caddy accepts this string with commas and whitespaces. Let's add a TODO to clean this up later.

flokli (dismissed):
Agree with the outcome, but no fan of the "comma-and-whitespace-separated URL", let's add a TODO for that.

vjuntunen (dismissed):

terraform/main.tf
flokli: I assume long term we only want to have one canonical domain, so this should just highlight the smell. If we start wanting to have multiple domains for an extended period of time, we'd need to make this a list, and do the templating to a comma-and-whitespace-separated list of urls shortly before we pass it to caddy.

henrirosten: Thanks, fixed in https://github.com/tiiuae/ghaf-infra/pull/216/commits/040ced3010f4175d244b4f07a74f5ad81ff5837e.

flokli (approved):

joinemm (approved):

PR #220, opened 2024-08-14
Increase blob storage create timeout

cazfi:
Set to 1h, doubling it from the default of 30m.


henrirosten:
Thanks @cazfi.

Looking at https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_blob#timeouts, shouldn't we also increase the `update` and `read` timeouts?

cazfi:
> Thanks @cazfi.
> 
> Looking at https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_blob#timeouts, shouldn't we also increase the `update` and `read` timeouts?

We could, but I've not yet seen any failures because of those ones.

henrirosten:
I would increase those other timeouts already in this PR, but you decide, I'm fine either way.

henrirosten (approved):

joinemm (approved):

PR #222, opened 2024-08-15
Github OAuth for ghaf-log grafana instance

joinemm:
Navigating to https://ghaflogs.vedenemo.dev now presents the user with "Sign in with Github" button. If the logged in Github user is part of the `tiiuae` org, and the `devenv-fi` or `phone` teams, they are allowed access to the grafana instance. ~~The default grafana login with username and password is no longer possible.~~

Grafana account is created for every user when they first log in: 
image

### DIFF
--- a/hosts/ghaf-log/configuration.nix
+++ b/hosts/ghaf-log/configuration.nix
@@ -35,8 +35,13 @@
       user-vunnyso
     ]);
 
-  # basic auth credentials generated with htpasswd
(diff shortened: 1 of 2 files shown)


joinemm:
Restored traditional login so the admin account can still be used.

vunnyso (approved):

cazfi (dismissed):

joinemm:
Reformatted with `nixfmt` now that it's the new standard.

PR #223, opened 2024-08-15
Invoke terraform init with upgrade option

henrirosten:
Make terraform emit an error message in case the azurerm version cached in local `.terraform` directory on the main terraform module does not match the version we require.

Make `terraform-init.sh` invoke the command `terraform init` with `-upgrage` option so the terraform plugin versions cached in local `.terraform` directory would be updated on updating the pinned terraform plugin versions - they are pinned in ghaf-infra flake devshell and should be updated when flake updates.

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -9,6 +9,12 @@ terraform {
   required_providers {
     azurerm = {
       source = "hashicorp/azurerm"
+      # Make sure this matches the version in ../nix/devshell.nix.
+      # This will not pin the version (it's pinned in devshell.nix),
+      # but makes terraform emit an error message in case the azurerm
+      # version cached in local .terraform directory for one reason
+      # or another does not match what is stated below.
+      version = "=3.85.0"
     }
     secret = {
       source = "numtide/secret"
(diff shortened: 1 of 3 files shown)


terraform/main.tf
```diff
@@ -9,6 +9,12 @@ terraform {
   required_providers {
     azurerm = {
       source = "hashicorp/azurerm"
+      # Make sure this matches the version in ../nix/devshell.nix.
+      # This will not pin the version (it's pinned in devshell.nix),
+      # but makes terraform emit an error message in case the azurerm
+      # version cached in local .terraform directory for one reason
+      # or antoher does not match what is stated below.
```
cazfi, line 16: s/antoher/another

cazfi (approved):

terraform/main.tf
henrirosten: Thanks, fixed in https://github.com/tiiuae/ghaf-infra/commit/b6abe2966b1e2871232b31fcc2c2c9690582b235

joinemm (approved):

mkaapu (approved):

cazfi (approved):

PR #226, opened 2024-08-16
HW test configuration file fix

vjuntunen:
Fix for lenovo-x1 testagent configuration.
There was naming issue of controlling right device button pressers.


azbeleva (approved):

henrirosten:
Maybe add a short explanation as to why the fix is needed?

henrirosten (approved):

PR #231, opened 2024-08-23
azure-nix-cache-proxy: send mime types

flokli:
We currently send application/octet-stream as a content-type, but we can use the correct mime types for NAR, NARInfo and nix-cache-info.

Using text/… for NARInfo and nix-cache-info prevents browsers from treating these URLs as a "Download" vs. just showing inline. Nix ignores them.

Nix uses application/x-nix-nar consistently, for compressed and uncompressed files.


henrirosten (approved):

henrirosten:
Deployed to dev

PR #236, opened 2024-08-31
jenkins-controller: Install Copy Artifact plugin

henrirosten:
Install Jenkins Copy Artifact plugin

### DIFF
--- a/hosts/azure/jenkins-controller/configuration.nix
+++ b/hosts/azure/jenkins-controller/configuration.nix
@@ -279,7 +279,7 @@ in
           "workflow-aggregator" "github" "timestamper" "pipeline-stage-view" "blueocean" \
           "pipeline-graph-view" "github-pullrequest" "antisamy-markup-formatter" \
           "configuration-as-code" "slack" "pipeline-utility-steps" "pipeline-build-step" \
-          "robot"
+          "robot" "copyartifact"
 
         # Disable initial install
         jenkins-cli ${jenkins-auth} groovy = < ${jenkins-groovy}


cazfi (approved):

vjuntunen (approved):

joinemm (approved):

hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -279,7 +279,7 @@ in
           "workflow-aggregator" "github" "timestamper" "pipeline-stage-view" "blueocean" \
           "pipeline-graph-view" "github-pullrequest" "antisamy-markup-formatter" \
           "configuration-as-code" "slack" "pipeline-utility-steps" "pipeline-build-step" \
-          "robot"
+          "robot" "copyartifact"
```
flokli, line 282: Isn't this not gonna expose us to the same problems when we were having artifacts on the local filesystem?

henrirosten: We are using the `copyartifact` plugin only to copy the test report (.html, .xml)[ from the testagent](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/ec77585b6b5e1f3c9e8f372712686719f9b36e04/utils.groovy#L185-L190) to the jenkins-controller VM. The 'permanent' storage of the test reports is the [same blob storage](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/ec77585b6b5e1f3c9e8f372712686719f9b36e04/utils.groovy#L191-L192) where the build results (images) are also stored.

flokli: Ok, it would make sense to add a comment describing this here and at the pipeline code making use of it, and a tmpfiles rule cleaning up old files (>2 days) from this directory, to avoid the disk to fill up.

PR #237, opened 2024-09-02
Add nixos configuration for the second testagent machine

joinemm:
- Refactored the testagent configuration.
- Updated rev of nixos-anywhere as it was having issues with most recent version of nix on my laptop.
- Removed unnecessary suspend blockers because now that there is no GDM the machine is headless and doesn't suspend anyway.
- Added testagent-dev config with disko module, and installed using nixos-anywhere (inv install).
- testagent-dev does not do anything yet and has no jenkins agent configured.

Note for future re-installs that the hardware has issues with kexec (was unable to fix), so I had to boot nixos installer iso before connecting to it with nixos-anywhere to avoid kexec.


vjuntunen (approved):

hosts/testagent-dev/configuration.nix
```diff
@@ -0,0 +1,93 @@
...
+    packages = [
+      brainstem
+      pkgs.usbsdmux
+    ];
+
+    # udev rules for test devices serial connections
+    extraRules = ''
+      SUBSYSTEM=="tty", ATTRS{idVendor}=="0403", ATTRS{idProduct}=="6001", ATTRS{serial}=="FTD1BQQS", SYMLINK+="ttyORINNX1", MODE="0666", GROUP="dialout"
+      SUBSYSTEM=="tty", ATTRS{idVendor}=="10c4", ATTRS{idProduct}=="ea71", ATTRS{serial}=="0642246B630C149011EC987B167DB04", ENV{ID_USB_INTERFACE_NUM}=="01", SYMLINK+="ttyRISCV1", MODE="0666", GROUP="dialout"
+    '';
```
henrirosten, line 78: I realize this is not a complete tested configuration yet, but I think it's worth mentioning that these udev rules needs to be tested in the new 'dev' setup. I would assume e.g. the 'serial' values stated above would not match the new 'dev' setup.

henrirosten (approved):

mkaapu:
> Note for future re-installs that the hardware has issues with kexec (was unable to fix), so I had to boot nixos installer iso before connecting to it with nixos-anywhere to avoid kexec.

We probably have some sort of documentation about our test HW and/or how to deploy these test machines. Would be worth to document these kind of exceptions and possible future development needs there.

mkaapu (approved):

PR #242, opened 2024-09-10
Fix for NUC serial port config

vjuntunen:
Fix for testagent NUC serial port configuration
and udev rule for that.


azbeleva (approved):

henrirosten (approved):
LGTM (assuming this has been tested).

PR #246, opened 2024-09-12
Add curl for Jenkins

mnokka-unikie:


hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -106,6 +106,7 @@ in
         zstd
         jq
         csvkit
+        curl
```
cazfi, line 109: What is this used for? Trying to figure out if it belongs here...

hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -106,6 +106,7 @@ in
...
       ]
       ++ [
         rclone # used to copy artifacts
```
cazfi, line 112: ...or here.

hosts/azure/jenkins-controller/configuration.nix
mnokka-unikie: Curl is used to digt out Jenkins infop when emailing possible bild breakers

cazfi (approved):

PR #250, opened 2024-09-18
Run multiple Jenkins agent services simultaneously

joinemm:
Run one agent for each hardware test device connected to the testagent-dev machine. This enables future work for running the tests in parallel by utilizing different nodes (jenkins side still in progress).

### DIFF
--- a/hosts/testagent-dev/configuration.nix
+++ b/hosts/testagent-dev/configuration.nix
@@ -4,12 +4,87 @@
   self,
   inputs,
   pkgs,
+  config,
   ...
 }:
 let
   # Vendored in, as brainstem isn't suitable for nixpkgs packaging upstream:
   # https://github.com/NixOS/nixpkgs/pull/313643
   brainstem = pkgs.callPackage ../../pkgs/brainstem { };
+
+  mkAgent =
+    device:
+    let
+      # Temporary url for development
+      controllerUrl = "https://ghaf-jenkins-controller-joonasrautiola.northeurope.cloudapp.azure.com";
+      workDir = "/var/lib/jenkins/agents/${device}";
+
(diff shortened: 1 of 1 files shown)


hosts/testagent-dev/configuration.nix
```diff
@@ -4,12 +4,85 @@
...
+      workDir = "/var/lib/jenkins/agents/${device}";
+
+      jenkins-connection-script = pkgs.writeScript "jenkins-connect.sh" ''
+        #!/usr/bin/env bash
+        set -eu
+        mkdir -p ${workDir}
+
+        # get agent.jar
+        if [ ! -f agent.jar ]; then
+          wget ${controllerUrl}/jnlpJars/agent.jar
```
cazfi, line 29: Put this in quotes even if URL is never expected to contain spaces.

You may want to check this entire shell script fragment with shellcheck.

hosts/testagent-dev/configuration.nix
```diff
@@ -4,12 +4,85 @@
...
+          nix
+          zstd
+          jq
+          csvkit
+          sudo
+          openssh
+          iputils
+          netcat
+          python3
+          wget
```
cazfi, line 73: Duplicate wget

cazfi (changes requested):

hosts/testagent-dev/configuration.nix
joinemm: good catch

hosts/testagent-dev/configuration.nix
joinemm: This and other similar cases fixed

vjuntunen (approved):

PR #253, opened 2024-09-18
az-jenkins-controller: jenkins-config needs mount

flokli:
Add RequiresMountsFor=/var/lib/jenkins to jenkins-config.service.

There's no point starting this before /var/lib/jenkins is mounted, as we read the admin password from there to interact with the jenkins API.

### DIFF
--- a/hosts/azure/binary-cache/configuration.nix
+++ b/hosts/azure/binary-cache/configuration.nix
@@ -42,7 +42,7 @@ in
   # Due to an implicit RequiresMountsFor=$state-dir, systemd
   # will block starting the service until this mounted.
   fileSystems."/var/lib/caddy" = {
-    device = "/dev/disk/by-lun/10";
+    device = "/dev/disk/azure/scsi1/lun10";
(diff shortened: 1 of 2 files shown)


flokli:
I figured out `/dev/disk/azure/scsi1/lun*` seems to exist even in cases where `/dev/disk/by-lun/*` symlinks failed to be created, so this might be a way out.

henrirosten (approved):
This seems very promising. I have not been able to reproduce the volume attachment issues with the changes from this PR. 

@flokli: do you think we could merge this?

flokli:
> This seems very promising. I have not been able to reproduce the volume attachment issues with the changes from this PR.
> 
> @flokli: do you think we could merge this?

I did not yet test this, as I didn't want to reprovision the machine while the support case is ongoing, which is why I marked it as draft. 

In case you did test however, and were not able to reproduce the issue anymore, we can merge it.

joinemm (approved):

PR #256, opened 2024-09-25
Add KV policy management

alextserepov:
- auto add jenkins controller to KV policy


henrirosten:
You might want to squash the commits into one, but otherwise tested that it works as expected.

henrirosten (approved):

alextserepov:
> You might want to squash the commits into one, but otherwise tested that it works as expected.

Yeah, fully agreed. I use "Squash and Merge" as default practice. Thanks for testing it!

PR #257, opened 2024-09-27
Add riscv to testagent-dev

vjuntunen:
Add riscv device configuration and
agent service in testagent-dev configuration.nix
Also udev rule for serial connection.

### DIFF
--- a/hosts/testagent-dev/configuration.nix
+++ b/hosts/testagent-dev/configuration.nix
@@ -139,11 +139,18 @@ in
     cpu.intel.updateMicrocode = true;
   };
 
-  # Enable Acroname USB Smart switch, as well as LXA USB-SD-Mux support.
-  services.udev.packages = [
-    brainstem
-    pkgs.usbsdmux
-  ];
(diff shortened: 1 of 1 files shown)


hosts/testagent-dev/configuration.nix
```diff
@@ -140,10 +140,23 @@ in
   };
 
   # Enable Acroname USB Smart switch, as well as LXA USB-SD-Mux support.
-  services.udev.packages = [
-    brainstem
-    pkgs.usbsdmux
-  ];
+  #services.udev.packages = [
```
cazfi, line 143: Remove this block completely instead of only commenting out.

cazfi (changes requested):

hosts/testagent-dev/configuration.nix
vjuntunen: fixed

joinemm (approved):

cazfi (approved):

PR #260, opened 2024-10-01
Retry post-build upload when azure blob is locked

joinemm:
Adds a for loop to the post-build-hook that retries the upload once in case it errors. This should stop the azure blob 423 locked error from completely failing our builds.

Not properly tested because I can't reproduce the issue in my infra, but the logic works as a standalone shell script.


cazfi (approved):

karim20230 (approved):

flokli (approved):
Looks simple enough. It might still fail if one upload takes more than 10 seconds, but we can increase the timeout if we hit that.

PR #269, opened 2024-10-09
Allow Jenkins to checkout from local git projects

mkaapu:
This change allows Jenkins to checkout changes from local git repositories.
This makes Jenkins project development faster when you only want to test changes made to Jenkins projects
without need to deploy whole ghaf-infra environment.

### DIFF
--- a/hosts/azure/jenkins-controller/configuration.nix
+++ b/hosts/azure/jenkins-controller/configuration.nix
@@ -136,6 +136,9 @@ in
       "-Dcasc.jenkins.config=${jenkins-casc}"
       # Increase the number of rows shown in Stage View (default is 10)
       "-Dcom.cloudbees.workflow.rest.external.JobExt.maxRunsPerJob=32"
+      # Allow checkouts done from local repositories in Jenkins
+      # Allows project development without need of constant deployments
+      "-Dhudson.plugins.git.GitSCM.ALLOW_LOCAL_CHECKOUT=true"
     ];
     # Configure jenkins job(s):
     # https://jenkins-job-builder.readthedocs.io/en/latest/project_pipeline.html


hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -136,6 +136,9 @@ in
       "-Dcasc.jenkins.config=${jenkins-casc}"
       # Increase the number of rows shown in Stage View (default is 10)
       "-Dcom.cloudbees.workflow.rest.external.JobExt.maxRunsPerJob=32"
+      # Allow checkouts done from local repositories in Jenkins
+      # Allows project development without need of constant deployments
+      "-Dhudson.plugins.git.GitSCM.ALLOW_LOCAL_CHECKOUT=true"
```
flokli, line 141: I'd feel more comfortable if this was not enabled for *all environments*, but triggered through some environment variable, file made available through cloud-init, ...

mkaapu: Overall, we could have common configuration for all environments, and some variations for separate environments. Or even git branches for separate environment.

I wanted to store this somewhere as I am going to delete my personal development VM soon.
If you feel too uncomfortable, we do not need to merge this.
Change does not really belong to current production environment.

flokli: Then let's put this in some of our markdown files, and not merge this.

henrirosten:
@mkaapu: Can we close this?

mkaapu:
> @mkaapu: Can we close this?

This is something which could be taken into use with priv/dev environments, when we can instantiate separate environments (or use different branches or whatever will be the solution).
But for now, it is ok to close this without merging/deploying.

PR #272, opened 2024-10-14
Increase image storage timeouts

henrirosten:
Increase image blob storage timeouts that were not increased by: https://github.com/tiiuae/ghaf-infra/pull/220

See: https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/storage_blob#timeouts


joinemm (approved):

karim20230 (approved):

cazfi (commented):
Maybe double each of them from their defaults, instead of setting all to an hour.

henrirosten:
> Maybe double each of them from their defaults, instead of setting all to an hour.

If you don't mind, I'd rather merge this with the current values to unblock people who are impacted by this.
If you want to, we could have a separate ticket to test and tweak with different values.

@cazfi: would this be fine?

PR #274, opened 2024-10-18
Automate creation of HW test nodes and fetching of secrets

joinemm:
Greatly simplifies the setup of testing environment or moving it to a different controller.

After this change, all you need to do to connect the test setup to a jenkins environment is to supply the controller url and admin password in `/var/lib/jenkins/jenkins.env`, and (re)start `setup-agents.service`. This service will then start the hardware agents.

Agent services creation moved to a separate file so it can be imported from all testagent configurations.

### DIFF
--- a/hosts/agents-common.nix
+++ b/hosts/agents-common.nix
@@ -1,0 +1,149 @@
+# SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+
+{ pkgs, inputs, ... }:
+let
+  # Vendored in, as brainstem isn't suitable for nixpkgs packaging upstream:
+  # https://github.com/NixOS/nixpkgs/pull/313643
+  brainstem = pkgs.callPackage ../pkgs/brainstem { };
+in
+{
+  services.udev.packages = [
+    brainstem
+    pkgs.usbsdmux
+  ];
+
+  environment.systemPackages =
+    [
+      inputs.robot-framework.packages.${pkgs.system}.ghaf-robot
+      brainstem
+    ]
+    ++ (with pkgs; [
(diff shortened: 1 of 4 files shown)


hosts/agents-common.nix
```diff
@@ -0,0 +1,148 @@
...
+  systemd.services =
+    let
+      # verifies that the environment file is properly configured
+      # and downloads agent.jar from the controller
+      jenkins-setup-script = pkgs.writeShellScript "jenkins-setup.sh" ''
+        if [[ ! -f jenkins.env ]]; then
+          echo "Please add jenkins controller details to $(pwd)/jenkins.env"
+          echo "CONTROLLER=" > jenkins.env
+          echo "ADMIN_PASSWORD=" >> jenkins.env
+          exit 1
```
henrirosten, line 46: After this, `/var/lib/jenkins/jenkins.env` is world-readable. We should create that file so only the owner has read and write access (600).

joinemm: fixed

hosts/agents-common.nix
```diff
@@ -0,0 +1,149 @@
...
+
+      # Helper function to create agent services for each hardware device
+      mkAgent =
+        device:
+        let
+          # opens a websocket connection to the jenkins controller from this agent
+          jenkins-connect-script = pkgs.writeShellScript "jenkins-connect.sh" ''
+            JENKINS_SECRET="$(
+              curl -L -s -u admin:$ADMIN_PASSWORD \
+              $CONTROLLER/computer/${device}/jenkins-agent.jnlp |
```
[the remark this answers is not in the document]
joinemm: Added `--proto =https` which will throw an error if the url is not using TLS.

henrirosten (approved):

PR #275, opened 2024-10-18
Refactor jenkins job builder configuration

joinemm:
- Refactored pipeline creation into a nix function to reduce duplicate code.
- Gave the pipelines more readable names.
- Added parallel hw test as it was missing.

### DIFF
--- a/hosts/azure/jenkins-controller/configuration.nix
+++ b/hosts/azure/jenkins-controller/configuration.nix
@@ -145,12 +145,13 @@ in
       enable = true;
       nixJobs =
         lib.mapAttrsToList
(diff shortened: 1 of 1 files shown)


mkaapu (dismissed):
LGTM

henrirosten:
At some point, I think we might not want to have all these pipelines configured on all deployments, but we can figure out how to support such selective config.

henrirosten (approved):

joinemm:
I changed this to set the display-name instead, so that references to pipelines in code still work.

henrirosten (approved):

PR #276, opened 2024-10-21
Ghaf infra new environments

henrirosten:
Support ghaf-infra deployments as outlined in https://ssrc.atlassian.net/wiki/spaces/SP/pages/1357250598/Ghaf-infra+Environments:

- Allow creating four types of different ghaf-infra instances: '**release**', '**prod**', '**dev**', '**priv**':
  - '**release**': ephemeral environment manually spun-up and destroyed for the purpose of making Ghaf SLSA L3 release builds. The ‘release’ environment does not use external builders, but spins-up the builder VMs together with the rest of the infra.
  - '**prod**': production environment provides the continuous services for Ghaf developers and testers: nix binary cache, as well as the CI environment. Prod environment(s) should run 24/7, minimizing downtime.
  - '**dev**':  development environment serves as the dev environment for ghaf-infra developers - to be able test infra changes before promoting them to ‘prod’. '**dev**' environment(s) are therefore otherwise similar to 'prod', except there's no uptime requirements. As an example, 'prod' and 'dev' share the binary cache storage and key.
  - '**priv**': personal environments for ghaf-infra development and testing.

This change also introduces changes to critical resource storage:
- Jenkins artifact storage is moved from environment-specific to persistent storage
- Separate persistent storage for 'release'
- Separate binary cache storage for 'priv' environments

In addition, the changes in this PR enable creating concurrent ghaf-infra instances of the same type. As an example, we could have two 'prod' deployments (with different names) running concurrently. Similarly, ghaf-infra developers can now have many 'priv' environments running concurrently.

Removes `terraform-playground.sh`: we should now use `terraform-init.sh` and the instructions in `terraform/README.md` to achieve the same functionality.


terraform/persistent/resources/main.tf
```diff
@@ -0,0 +1,105 @@
...
+    }
+    secret = {
+      source = "numtide/secret"
+    }
+  }
+}
+
+################################################################################
+
+terraform {
```
karim20230, line 23: Duplicate `terraform` blocks! should be combined into a single block to avoid confusion.

terraform/persistent/resources/main.tf
```diff
@@ -0,0 +1,105 @@
...
+
+locals {
+  # Raise an error if workspace is 'default',
+  # this is a workaround to missing asserts in terraform:
+  assert_workspace_not_default = regex(
+    (terraform.workspace == "default") ?
+  "((Force invalid regex pattern)\n\nERROR: workspace 'default' is not allowed" : "", "")
+
+  # Sanitize workspace name:
+  ws = substr(replace(lower(terraform.workspace), "/[^a-z0-9]/", ""), 0, 16)
```
karim20230, line 49: I think `/` is not necessary!

terraform/persistent/resources/main.tf
```diff
@@ -0,0 +1,105 @@
...
+
+# Current signed-in user
+data "azurerm_client_config" "current" {}
+
+
+################################################################################
+
+# Resources
+
+# secret_resouce must be created on import, e.g.:
```
karim20230, line 65: should be "secret_resource"

terraform/main.tf
```diff
@@ -113,16 +111,14 @@ locals {
...
-      ]
-      ext_builder_keyscan = ["builder.vedenemo.dev", "hetzarm.vedenemo.dev"]
+      binary_cache_public_key = "priv-cache.vedenemo.dev~1:FmJGfAkx+2fhqpzHGT/V3M35VcPm2pfkCuiTo8xQD0A="
+      binary_cache_url        = local.binary_cache_url_common
+      ext_builder_machines    = local.ext_builder_machines
+      ext_builder_keyscan     = local.ext_builder_keyscan
     }
     dev = {
+      builder_sshkey_id       = "ext"
+      persistent_id           = "prod"
```
karim20230, line 121: maybe incorrect "prod" persistent_id  in dev environment ?

karim20230 (commented):
quite large changes -hoping tested well!

terraform/persistent/resources/main.tf
joinemm: it is in fact necessary https://developer.hashicorp.com/terraform/language/functions/replace

henrirosten: You mean the two `/` in the regex?
It is required. Please see the documentation: https://developer.hashicorp.com/terraform/language/functions/replace:

> _If substring is wrapped in forward slashes, it is treated as a regular expression, ..._

That's exactly what we want, the string to be treated as a regex.

terraform/persistent/resources/main.tf
henrirosten: Fixed in https://github.com/tiiuae/ghaf-infra/commit/eb4eb5ae550bcd4ab71ac97ea1733ba79bb3490f

terraform/persistent/resources/main.tf
henrirosten: Fixed in https://github.com/tiiuae/ghaf-infra/commit/eb4eb5ae550bcd4ab71ac97ea1733ba79bb3490f

terraform/persistent/resources/main.tf
karim20230: I think you are correct! it is treated as a regex..

terraform/main.tf
henrirosten: That's intentional. As outlined in https://ssrc.atlassian.net/wiki/spaces/SP/pages/1357250598/Ghaf-infra+Environments and also in this PR's description, we want the `dev` and `prod` to share the same binary cache storage and nix signing key.

karim20230 (approved):

henrirosten:
> quite large changes -hoping tested well!

@karim20230 : thanks for the review. I also struggled as to how to split this in smaller, more digestible pieces. I couldn't figure out how to split the first two commits that are required align with the new setup. Other commits in this PR are smaller, hopefully easier to understand.

henrirosten:
Rebased on latest main

joinemm (approved):

terraform/terraform-init.sh
```diff
@@ -93,146 +117,126 @@ azure_location_to_shortloc () {
...
+    STATE_ACCOUNT="ghafinfra0state${SHORTLOC}"
+    PERSISTENT_RG="ghaf-infra-0-persistent-${SHORTLOC}"
+    echo "[+] Using state '$STATE_RG'"
+    echo "[+] Using persistent '$PERSISTENT_RG'"
+    echo "storage_account_rg_name=$STATE_RG" >"$MYDIR/.env"
+    echo "storage_account_name=$STATE_ACCOUNT" >>"$MYDIR/.env"
+    echo "persistent_rg_name=$PERSISTENT_RG" >>"$MYDIR/.env"
 }
 
 init_state_storage () {
```
flokli, line 136: This can be misinterpreted as only doing `terraform init`, while it's actually also doing an apply.

Consider renaming this (and the echo) to something more descriptive, like `apply_state_storage`

flokli: This can be done in a followup, shouldn't block this from landing!

flokli (approved):

PR #279, opened 2024-10-23
Don't ban office ip address

joinemm:


vjuntunen (dismissed):

karim20230 (dismissed):

cazfi:
I hate it when we have to hardcode internal details of our own network to what is supposed to be code reusable also in other installations.
At least add a comment stating that this needs to be changed on 3rd party installations.

henrirosten (approved):

joinemm:
> I hate it when we have to hardcode internal details of our own network to what is supposed to be code reusable also in other installations.

There are many places in the infra already where our internal ip addresses or specific hardware are referenced, but yes I can add a comment explaining what this address is.

henrirosten (approved):

PR #280, opened 2024-10-24
Increase post-build upload attempts

henrirosten:
Retry post-build-hook upload three times at 10 second interval before giving up.

Before this change, it would retry two times. However, there have been some cases where the upload failed the second time also which would fail the build before this change.


henrirosten:
Pushed version https://github.com/tiiuae/ghaf-infra/commit/e012f2f0540fd2d8f458e955f233d4f6d3abfd63, let's not rewrite the post-build hook but go with minimal changes, also being a bit more verbose on failure.

joinemm (approved):

PR #284, opened 2024-10-30
Start using build4 as x86 external builder in Azure infra

henrirosten:
- Move user-remote-build out of developers.nix: explicitly import user-remote-build on hosts that may be used from Azure jenkins-controller initiated 'external' builds.
- Start using `remote-build@build4.vedenemo.dev` as external x86 builder from jenkins-controller
- We need to keep remote-build user on build3 until all currently deployed ghaf-infra jenkins-controller instances that make use of external builders, have been migrated to a version of ghaf-infra including the changes from this PR.

### DIFF
--- a/hosts/builders/build3/configuration.nix
+++ b/hosts/builders/build3/configuration.nix
@@ -23,7 +23,8 @@
       user-themisto
       user-ktu
       user-avnik
-      user-github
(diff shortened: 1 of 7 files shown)


karim20230 (approved):

henrirosten:
Deployed to: build3, build4, hetzarm

PR #287, opened 2024-10-31
Update main readme Deploy by Using deploy-rs

karim20230:
Update main README Using deploy-rs  as an alternative to using `tasks.py` and `invoke deploy`


joinemm:
You should just to the file in `docs/deploy-rs` instead of copying its contents, now there's two places with same content to keep in sync

henrirosten:
I think we should also do a bigger cleanup for the main README.md (rewrite?), it now talks about ghafhydra and other hosts no longer relevant.

I would also like to remove the `task.py`, or at least parts of it now replaced with `deploy-rs`.

karim20230:
> I think we should also do a bigger cleanup for the main `README.md` (rewrite?), it now talks about `ghafhydra` and other hosts no longer relevant.
> 
> I would also like to remove the `task.py`, or at least parts of it now replaced with `deploy-rs`.

I completely agree with you! rewriting our main README is a great idea!

PR #289, opened 2024-11-01
Skip x86 builder image unless builder VMs are created

henrirosten:
- Do not upload x86 builder vhd image to azure if the infrastructure will not create the x86 builder VM.
- Add instructions to terrafrom/README.md to work around an azurerm issue this change exposes.


vjuntunen (approved):

mnokka-unikie (approved):
go

joinemm (approved):

PR #298, opened 2024-11-12
Nix trusted public key in persistent

henrirosten:
- Store nix trusted public key in persistent storage
- Read the public key from the persistent storage instead of hard-coding the expected value in the terraform main module
- Change `test-deployment.sh` to support changed configuration: expected trusted public key values for eun location are pre-defined in the script - it also now supports providing the trusted public key from the command line with a new option `-p`


joinemm (approved):

mnokka-unikie (approved):
go

PR #304, opened 2024-11-15
 Start using nix-fast-build in github action CI

henrirosten:
- Move pycodestyle, pylint, black, and format checks from `task.py` to ghaf-infra flake `.#checks`
- Remove all redundant checks from `task.py`
- Introduce `nix-fast-build.sh` helper to run relevant flake builds with [nix-fast-build](https://github.com/Mic92/nix-fast-build) to execute all evaluation/build operations completely on the remote builder. Add relevant dependencies to nix `devshell`.
- Start using `nix-fast-build.sh` from github action workflow `test-ghaf-infra.yml`: run all checks and build relevant x86 and aarch targets on remote. This adds some new checks too: the workflow now also builds all azure nixosConfigurations, as well as runs the `.#checks` on aarch64.

Note: after this change, cachix cache on the `test-ghaf-infra.yml` workflow  is used only to cache the ghaf-infra `devshell` dependencies. This is to make the `test-ghaf-infra.yml` faster by not having to download the build closure to the github-hosted runner. Rationale is that local builds of all the ghaf-infra nixosConfigurations are likely rare, and when required, `nix-fast-build.sh` can be used locally too (in which case the remote builder's local nix store would be used as a 'shared cache').

Example run of `test-ghaf-infra.yml` workflow with the changes from this PR: https://github.com/henrirosten/ghaf-infra/actions/runs/11854339487

### DIFF
--- a/nix/checks.nix
+++ b/nix/checks.nix
@@ -1,19 +1,20 @@
 # SPDX-FileCopyrightText: 2022-2024 TII (SSRC) and the Ghaf contributors
 # SPDX-License-Identifier: Apache-2.0
-_: {
+{ self, ... }:
+{
   perSystem =
     { pkgs, ... }:
     {
       checks = {
         reuse = pkgs.runCommandLocal "reuse-lint" { buildInputs = [ pkgs.reuse ]; } ''
-          cd ${../.}
+          cd ${self.outPath}
           reuse lint
           touch $out
         '';
         pycodestyle =
           pkgs.runCommandLocal "pycodestyle" { nativeBuildInputs = [ pkgs.python3.pkgs.pycodestyle ]; }
             ''
-              cd ${../.}
+              cd ${self.outPath}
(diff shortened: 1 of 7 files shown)


nix/checks.nix
```diff
@@ -10,6 +10,30 @@ _: {
           reuse lint
           touch $out
         '';
+        pycodestyle =
+          pkgs.runCommandLocal "pycodestyle" { nativeBuildInputs = [ pkgs.python3.pkgs.pycodestyle ]; }
+            ''
+              cd ${../.}
```
joinemm, line 16: nitpick but `${self}` can be used to refer to the flake root in more portable way than relative path

joinemm (approved):

nix/checks.nix
henrirosten: Fixed in https://github.com/tiiuae/ghaf-infra/pull/304/commits/fda27a0da3c970c229892cfbc7159e7aaa21cd99

joinemm (approved):

PR #320, opened 2024-12-04
Terragrunt setup, DNS delegation for az.vedenemo.dev

flokli:
This sets up a DNS zone for az.vedenemo.dev, and allows different environments to manage their own DNS zones, one subdomain per environment.


terraform/prod-eun/README.md
```diff
@@ -0,0 +1,20 @@
...
+It should also make it much easier to do a granular destruction of a "leaf"
+state.
+
+It does not use Terraform Workspaces, individual instantiations have their own
+files in this repository, making it possible to deviate or roll out changes in a
+controlled matter.
+
+In case the experiment proves successful, the goal is to slowly migrate
+everything from prod-eun (and other environments) to this approach, creating
+sibling directories.
```
flokli, line 20: I want to add a quick start and link to some more documentation, keeping it draft for now.

flokli: Done

henrirosten:
Thanks @flokli

I agree there are problems with the current workspace-based approach, anything we can do to improve the workflow or make our terraform usage a bit more industry standard is very much welcome.

- Can you reference a good source of documentation that would help see a bigger picture of how this should be used and how the layout would eventually end-up?
- Am I reading this correctly, that we could merge this now, and then slowly move other parts of the infra and different deployments to this new layout too?
- You could try converting your personal ghaf-infra environment to this new layout - then we could probably better see what are all the problems that will arise. 

Also, can you reference a good source of documentation that would help others also see the bigger picture of how this should be used and how the layout would eventually end-up?

henrirosten (approved):

flokli:
> Thanks @flokli
> 
> I agree there are problems with the current workspace-based approach, anything we can do to improve the workflow or make our terraform usage a bit more industry standard is very much welcome.
> 
>     * Can you reference a good source of documentation that would help see a bigger picture of how this should be used and how the layout would eventually end-up?

The bigger picture would be that the different components would be terraform modules instantiated from the different environment, but explicitly written down for each environment, and things like persistent volumes passed from one Terraform State to another via the [inputs mechanism](https://terragrunt.gruntwork.io/docs/features/inputs/).

I'm not aware of any public repo using terragrunt to showcase this more, but there's some generic documentation describing some of the DRY practices: https://terragrunt.gruntwork.io/docs/features/keep-your-terragrunt-architecture-dry/

>     * Am I reading this correctly, that we could merge this now, and then slowly move other parts of the infra and different deployments to this new layout too?

Yes. The fact we now have an additional DNS Zone managed this way won't cause any problems with the existing Terraform states, so this can be merged.

>     * You could try converting your personal ghaf-infra environment to this new layout - then we could probably better see what are all the problems that will arise.

Yes, starting with my personal environment was the next step I had in mind, and this should make this a bit less abstract, and "flesh out some calling conventions"

> Also, can you reference a good source of documentation that would help others also see the bigger picture of how this should be used and how the layout would eventually end-up?

I'd assume more directories to appear in the `terraform/` directory, in addition to the `prod-eun` and `vedenemo` ones, one for each environment. The `terraform/*.tf` files will go away, once everything is migrated. We might then "factor out" some common calling code, if they're the same across environments to reduce duplication, but that's a second step once the code has been moved to the new layout.

PR #321, opened 2024-12-05
Update infra to NixOS 24.11

joinemm:
- `jenkins-job-builder` is currently broken, so using the old version for now. Issue has been filed: https://github.com/NixOS/nixpkgs/issues/362054 and fix is coming.

- rclone 1.68.2 breaks our pipelines, archival step results in an error (path truncated):
  ```
  2024/12/09 12:39:45 CRITICAL: Failed to create file system for ":webdav:/ghaf-main-pipeline/build_5-commit_...": 
  read metadata failed: 308 Permanent Redirect
  ```
  Reverting to 1.66 with our webdav patches fixes the issue, but this should be looked into @flokli 

- Migration steps for azure environment: run `./terraform-init -w` again before applying.

### DIFF
--- a/services/rclone-http/default.nix
+++ b/services/rclone-http/default.nix
@@ -9,7 +9,6 @@
 with lib;
 let
   cfg = config.services.rclone-http;
-  rclone = pkgs.callPackage ../../pkgs/rclone { };
 in
 {
   options.services.rclone-http = {
@@ -56,7 +55,7 @@ in
         EnvironmentFile = "/var/lib/rclone-http/env";
         ExecStart = concatStringsSep " " (
           [
-            "${rclone}/bin/rclone"
+            "${pkgs.rclone}/bin/rclone"
             "serve"
             cfg.protocol
           ]
--- a/flake.lock
+++ b/flake.lock
@@ -206,6 +206,22 @@
         "type": "github"
       }
     },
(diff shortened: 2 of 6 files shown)


services/rclone-http/default.nix
```diff
@@ -9,7 +9,6 @@
 with lib;
 let
   cfg = config.services.rclone-http;
-  rclone = pkgs.callPackage ../../pkgs/rclone { };
```
flokli, line 12: Upstream rclone >= v1.68.0 should have all patches we need and have. `pkgs/rclone/*` can be removed, it only contained the override and vendored patches.

flokli: Ah, I see, we call this explicitly elsewhere, ugh.

I'd like to see this cleaned up in the future. Can you provide some more detailed logs how exactly this broke the uploads?

joinemm: Yes we still need the old patched rclone in the jenkins controller. Unfortunately the log I shared is the only error I got out of jenkins so I have no idea what is going wrong there.

This can be reproduced by running any pipeline in a jenkins controller instance that has rclone >= 1.68.0 installed. The archival step will fail.

henrirosten (approved):

cazfi (approved):

PR #330, opened 2024-12-11
ghaf-webserver: Add gala uploaders

cazfi:


hosts/ghaf-webserver/gala_uploaders.nix
```diff
@@ -0,0 +1,50 @@
...
+          openssh.authorizedKeys.keys = keys;
+
+          isNormalUser = true;
+          extraGroups = [ groupName ];
+        }
+      ) uploaders
+    );
+  };
+
+  nix.settings.trusted-users = [ "@${groupName}" ];
```
joinemm, line 49: I don't know how the gala packages are uploaded but do the users need to be trusted by nix?

cazfi: No, they only need to be able to scp. Fixed.

karim20230 (approved):
LGTM

joinemm (approved):

PR #352, opened 2025-01-22
az-jenkins-controller: handle auth by delegating to oauth2-proxy

flokli:
This configures caddy to delegate authentication to a deployment of
    oauth2-proxy, which will redirect the user to GitHub for authentication.
    
    We peek at the user profile, and in case they're part of the tiiuae
    organization, let them in.
    
    As GitHub redirects back to the application after the users consent, we
    need to have one GitHub application per Jenkins deployment.
    Said GitHub OAuth application also needs to have access to see
    membership in the tiiuae application.
    
    This will become easier in the future by adding a "trampoline" IdP
    like Dex in between, allowing us to manage individual clients and their
    redirect URIs locally, with only the Dex URL needing to be configured
    in GitHub.


flokli:
Depends on #361. Might also need changes to deal with authentication of the agents.

joinemm:
closing in favor of https://github.com/tiiuae/ghaf-infra/pull/400

PR #353, opened 2025-01-23
Prevent workspace name conflict

henrirosten:
On initializing 'workspace persistent' data, we need to check the user-defined WORKSPACE name does not overlap with one of the [hardcoded (and thus reserved)](https://github.com/tiiuae/ghaf-infra/blob/2bb4bc6b3ceb3991eccc9df12593312b6e636b77/terraform/terraform-init.sh#L196) persistent workspace names.

This change adds a check that aborts `terraform-init.sh` in case user-defined WORKSPACE overlaps.


fayadf (approved):
Looks good, tested on uae prod.

PR #354, opened 2025-01-27
Flake update

joinemm:
- Jenkins is updated to fix a CVE https://github.com/NixOS/nixpkgs/pull/360869 and also for better plugin compability
- jenkins-job-builder is now fixed so we don't have to use the 24.05 version.
- Coverity had a hash mismatch -> updated hash.
- Most recent nix formatter behaves slightly differently so had to do a reformat.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -50,11 +50,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1733168902,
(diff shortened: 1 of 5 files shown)


vjuntunen (approved):

henrirosten (approved):
Assuming this has been tested

PR #356, opened 2025-01-30
Jenkins plugin setup improvements

joinemm:
- `jenkinsPlugins2nix` is used to generate a set of jenkins plugins to install, complete with pinned versions and hashes.
- Installed `lockable-resources` plugin so we can lock resources in pipelines.
- Jenkins CASC changes
  - Removed default values, only keeping what we actually change. This makes it much more readable.
  - timestamper plugin configured to enable timestamps in every job.
  - pipeline graph now shows up on build page without extra clicking.
  - more executors.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -87,6 +87,14 @@
         treefmt-nix.follows = "treefmt-nix";
       };
     };
+
+    jenkinsPlugins2nix = {
+      url = "github:Fuuzetsu/jenkinsPlugins2nix";
+      inputs = {
+        flake-compat.follows = "flake-compat";
+        flake-utils.follows = "flake-utils";
+      };
+    };
   };
 
   outputs =
--- a/hosts/azure/jenkins-controller/plugins.toml
+++ b/hosts/azure/jenkins-controller/plugins.toml
@@ -1,0 +1,16 @@
+commons-compress-api:1.26.1-2 # 1.27 is not compatible with jenkins 2.479.2
+configuration-as-code
+workflow-aggregator
+github
+slack
+timestamper
+blueocean
+pipeline-stage-view
+pipeline-graph-view
+github-pullrequest
+antisamy-markup-formatter
+pipeline-utility-steps
+pipeline-build-step
+copyartifact
+robot
+lockable-resources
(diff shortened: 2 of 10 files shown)


henrirosten (approved):

flake.nix
```diff
@@ -87,6 +87,14 @@
         treefmt-nix.follows = "treefmt-nix";
       };
     };
+
+    jenkinsPlugins2nix = {
+      url = "github:Fuuzetsu/jenkinsPlugins2nix";
+      inputs = {
+        flake-compat.follows = "flake-compat";
+        flake-utils.follows = "flake-utils";
```
flokli, line 95: is it possible to follow the same nixpkgs as the rest, so we don't pull in another nixpkgs instance?

hosts/azure/jenkins-controller/plugins.toml
```diff
@@ -0,0 +1,16 @@
+commons-compress-api:1.26.1-2 # 1.27 is not compatible with jenkins 2.479.2
```
flokli, line 1: This is not valid toml. maybe just call it `plugins.txt`?

flake.nix
joinemm: unfortunately not, `jenkinsPlugins2nix` doesn't build with the most recent nixpkgs

hosts/azure/jenkins-controller/plugins.toml
joinemm: I changed the extension to toml to get syntax highlighting for comments, maybe a stupid reason but it works :smile:

flokli: I mean, technically we could even write this out entirefly into `prefetch-plugins.sh` too, explicitly write out the `jenkinsPlugins2nix` invocation, and add comments there.

This is not a native format `jenkinsPlugins2nix` format, and we only use it in this one place. Then there's no also no need to debug how `jenkinsPlugins2nix` is gonna be called, because it's written out literally.

But that's my personal opinion, there's no need to block this PR for any of this. Happy to deal with it in a followup.

PR #399, opened 2025-04-03
Allow using more than one testagent per controller

joinemm:
Adds all three testsets as nodes to each controller (dev, prod and release). Any combination of these can connect. Builds can use both at the same time with the device label.

Refactored testagents config to be more consistent, and moved the agent connection script to its own module with an option.

### DIFF
--- a/hosts/azure/jenkins-controller/jenkins-casc.yaml
+++ b/hosts/azure/jenkins-controller/jenkins-casc.yaml
@@ -15,34 +15,114 @@ jenkins:
       disableSyntaxHighlighting: false
 
   nodes:
+    # release testagent
   - permanent:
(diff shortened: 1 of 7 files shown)


hosts/azure/jenkins-controller/jenkins-casc.yaml
vjuntunen, line 1: Are this all different agents for target using same label? So execution can end-up randomly any available one that have same labelString?

joinemm: correct

vjuntunen (approved):

henrirosten (approved):

PR #400, opened 2025-04-04
Jenkins github authentication using OAuth2 proxy and central auth server

joinemm:
- Jenkins is behind OAuth2 proxy, which delegates auth to `dex` running on `auth.vedenemo.dev` (hetzner server)
- Testagents connect through SSH to grab their websocket secret.
- Auth passes back username and github teams, which are mapped to jenkins groups for matrix permissions.
- Auth client secret is stored as sops secret for terraform to decrypt during planning.
- Flake inputs updated, Jenkins updated, Jenkins plugins updated.


hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -392,6 +392,32 @@ in
...
+          path /github-webhook /github-webhook/*
+          path /instance-identity /instance-identity/*
+          path /jnlpJars /jnlpJars/*
+          path /jwt-auth /jwt-auth/*
+          path /metrics /metrics/*
+          path /reload-configuration-as-code /reload-configuration-as-code/*
+          path /static-files /static-files/*
+          path /subversion /subversion/*
+          path /wsagents /wsagents/*
+        }
```
henrirosten, line 541: Can you reference the source of this list of endpoints that should remain unauthenticated?
My preference would be to authenticate all endpoints, unless there's a specific use-case in our infra that requires the endpoint to remain public.

joinemm: Listed here along with the options for reverse proxy auth plugin https://ghaf-jenkins-controller-jrautiola.northeurope.cloudapp.azure.com/manage/configureSecurity/

henrirosten: I notice 'whoAmI' is missing from the above list vs. what is mentioned in the link you sent.

I'm fine with this list now, but maybe we should revisit this soon to try to minimize the list and require authentication for all endpoints unless our use-case needs it public.

joinemm: yeah, whoAmI was pretty useless when it didn't include the autheticated user's information. From this list we only really need `/wsagents` and `/jnlpJars`, but since a list was provided I simply used it.

henrirosten:
Overall this seems fine to me.

You might want to add some documentation too (in this PR or later if you prefer), explaining how the authentication now works perhaps using jenkins-controller as an example. This would make it easier to have an overview in mind when one tries to understand the related configurations introduced in this PR.

henrirosten (dismissed):

hosts/azure/jenkins-controller/configuration.nix
joinemm: I reduced this list to only paths we need to stay unauthenticated

henrirosten (approved):

scripts/connect-testagent.sh
```diff
@@ -12,125 +12,110 @@ MYNAME=$(basename "$0")
...
 }
 
 # shellcheck disable=SC2086 # intended word splitting of $OPT_a
 # shellcheck disable=SC2029 # intended client side expansion of $CONTROLLER
 write_jenkins_env() {
-    if ! ssh $OPT_a "\
-        sudo bash -c 'echo 'CONTROLLER=$CONTROLLER' > /var/lib/jenkins/jenkins.env'
-        sudo bash -c 'echo 'ADMIN_PASSWORD=$ADMIN_PASSWORD' >> /var/lib/jenkins/jenkins.env'
+	if ! ssh $OPT_a "\
+        sudo bash -c 'echo 'CONTROLLER=$OPT_c' > /var/lib/jenkins/jenkins.env'
```
alextserepov, line 94: Could you please add $OPT_c validation before use? I'm not 100% certain, but it feels we might have potential for command injection here.

alextserepov (changes requested):
This is looking good. I agree with Henri's comment that adding some documentation would be helpful. Please include input validation as well, and I believe we’ll be all set.

scripts/connect-testagent.sh
joinemm: actually I'm going to remove this whole script as it's no longer needed

joinemm:
I'm going to write some documentation into this PR

scripts/connect-testagent.sh
alextserepov: Ok. That's even better. :)

docs/jenkins-authentication.md
```diff
@@ -0,0 +1,17 @@
+<!--
+SPDX-FileCopyrightText: 2022-2025 TII (SSRC) and the Ghaf contributors
+SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+# Jenkins authentication
+
+The Jenkins authentication setup relies on a few pieces working together:
+
+- Github app installed in the tiiuae org
```
henrirosten, line 9: Looking at the image you added, is this github app one of the boxes (which one)?
Is there any more details we could share about the github app?

joinemm: that would be the box labeled github login. Added some more details.

henrirosten (approved):

alextserepov (approved):

PR #422, opened 2025-04-22
Add cachix-watch-store to build1 and hetzarm

henrirosten:
Main changes:
- Add host keys to `build1` and `build2`
- Add [cachix-watch-store service](https://search.nixos.org/options?channel=unstable&show=services.cachix-watch-store.enable&from=0&size=50&sort=relevance&type=packages&query=cachix-watch-store) and cachix authentication token for hosts `build1` and `hetzarm` to start populating [ghaf-dev](https://app.cachix.org/organization/tiiuae/cache/ghaf-dev) cachix cache
 
After this PR is deployed, `cachix-watch-store` service on hosts `build1` and `hetzarm` starts pushing all newly built nix store paths to cache https://ghaf-dev.cachix.org. If this turns out to be too much, we might later want to limit the pushes, but I'd propose we start with this.

### DIFF
--- a/.sops.yaml
+++ b/.sops.yaml
@@ -17,6 +17,8 @@ keys:
   - &testagent-prod age12nrv5a9rk9vqvx2tqvghn4kt9ps6gdszmmynhjegl2ewefkh03fsexuy9y
   - &testagent-dev age1qjhxuh80tg2vq32kmwu2ne4vqvd8q2up7css30x0yefkrhq9jd0sxju3fa
   - &testagent-release age1pkd7crz3c0axuy4nnesjjzmklmqptx9fq5v8ndjevfgr5lqg8cgskhfyt7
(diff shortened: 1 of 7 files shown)


.sops.yaml
```diff
@@ -60,7 +62,18 @@ creation_rules:
       - *testagent-release
       - *jrautiola
       - *vjuntunen
```
joinemm, line 64: you've removed yourself from the release agent secrets (unintentionally?)

joinemm (changes requested):

.sops.yaml
henrirosten: Thanks, good catch. That was certainly unintentional. Fixed in https://github.com/tiiuae/ghaf-infra/commit/ac522b3d6ae4dfd6d76c1f3d0f894072d6fc3b33.

joinemm (approved):

henrirosten:
Deployed to: `build1`, `build2`, and `hetzarm`.

PR #430, opened 2025-05-02
Improve pre-commit hooks, checks and formatters

joinemm:
- Custom git commit message check script is replaced with `gitlint`.
- `treefmt` is now ran as pre-commit check.
- Moved all formatters from separate checks to treefmt options. `black` is replaced with `ruff format`. `nixfmt` defaults to rfc style now.
- Added `actionlint` to lint github actions workflows.


vjuntunen (approved):

henrirosten:
This is great, thanks @joinemm.
I believe we should also now remove this section from README.md: https://github.com/tiiuae/ghaf-infra/blob/496ada9bbd82355df6f98a19e4c2df3f0fddc4ba/README.md?plain=1#L117-L122

joinemm:
@henrirosten Good catch, removed.

PR #433, opened 2025-05-05
Add monitoring service to controller

vjuntunen:
Add monitoring service to Azure Jenkins comtrollers (dev, prod and vjuntunen)
and job for that on monitoring server

### DIFF
--- a/hosts/monitoring/configuration.nix
+++ b/hosts/monitoring/configuration.nix
@@ -63,6 +63,44 @@ in
       "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPc04ZyZ7LgUKhV6Xr177qQn6Vf43FzUr1mS6g3jrSDj";
   };
 
+  systemd.services.populate-jenkins-known-hosts = {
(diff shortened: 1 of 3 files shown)


hosts/monitoring/configuration.nix
```diff
@@ -249,6 +303,31 @@ in
...
       }
+      {
+        job_name = "Azure-Jenkins-monitoring";
+        # proxy the requests through our ssh tunnel
+        proxy_url = "http://127.0.0.1:8888";
+        static_configs = [
+          {
+            targets = [ "ghaf-jenkins-controller-vjuntunen.northeurope.cloudapp.azure.com:9100" ];
+            labels = {
+              machine_name = "Jenkins vjuntunen controller";
```
joinemm, line 314: we probably don't want to monitor private environments as they get taken down for long periods of time

vjuntunen: It is there just for some testing/iprovements... If we catch some errors etc.. Iwill remove it later yes.

joinemm (approved):

PR #441, opened 2025-05-09
rclone: Get from nixpkgs-24.11

cazfi:
Remove special case that rclone is fetched
from nixpkgs-24.05.

Interdependent with
https://github.com/tiiuae/ghaf-jenkins-pipeline/pull/155


hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -445,7 +440,7 @@ in
       RuntimeDirectory = "rclone-http";
       EnvironmentFile = "/var/lib/rclone-jenkins-artifacts/env";
       ExecStart = lib.concatStringsSep " " [
-        "${rclone}/bin/rclone"
+        "rclone"
```
joinemm, line 443:
```suggestion
        "${pkgs.rclone}/bin/rclone"
```
rclone is not in PATH for this service. either use from pkgs or add to PATH

hosts/azure/jenkins-controller/configuration.nix
```diff
@@ -478,7 +473,7 @@ in
...
```
joinemm, line 476: same here

joinemm (changes requested):

hosts/azure/jenkins-controller/configuration.nix
cazfi: Fixed in latest version

henrirosten (approved):

joinemm (approved):

PR #459, opened 2025-05-28
Use disk/by-id/ instead of labels in disko configuration

joinemm:
Updated most disk configurations with the correct disk ids instead of using labels like /dev/sda which can change on reboot. Left out himalia and binarycache since they aren't online.

### DIFF
--- a/hosts/builders/build1/configuration.nix
+++ b/hosts/builders/build1/configuration.nix
@@ -23,6 +23,13 @@
 
   # build1 specific configuration
 
+  disko.devices.disk = {
+    sda.device = "/dev/disk/by-id/ata-DELLBOSS_VD_74c47b68af530010";
+    sdb.device = "/dev/disk/by-id/scsi-362cea7f07374e7002785cfaa0d8881ae";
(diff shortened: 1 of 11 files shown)


cazfi (approved):

henrirosten:
Test faillure is unrelated to this change:

```
ssh: connect to host build1.vedenemo.dev port 22: Connection timed out
WARNING:nix_fast_build:build  exited with 255
INFO:nix_fast_build:builds: 0, uploads: 0, downloads: 0
parallel: This job failed:
fast_build nixosConfigurations.ghaf-coverity.config.system.build.toplevel
```

I assume ssh connection to build1.vedenemo.dev remote builder timeouts due to too much load on that host.
We might have to tweak the [parallel builds in nix-fast-build.sh ](https://github.com/tiiuae/ghaf-infra/blob/10b631de511ecda0a0f18e63027ad333f8256962/scripts/nix-fast-build.sh#L257-L264) to not overload the builder. Anyway, that would be a separate task, not related to the changes in this PR.

PR #469, opened 2025-06-11
Increase ficolo builders max jobs to 32

joinemm:


joinemm:
deployed to ficolo builders

PR #489, opened 2025-06-24
Hetzci update: jenkins plugins

henrirosten:
- Update hetzci-prod/dev jenkins plugins
- Additionally, add dev, prod, and uae-dev testagent users to hetzci-prod/dev


henrirosten:
Deployed to hetzci-prod and hetzci-dev

PR #505, opened 2025-07-01
hetzci: Run perf test nightly

henrirosten:
- Add pipeline 'ghaf-nightly-perftest' with same functionality as [ghaf-perftest-pipeline](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/main/ghaf-perftest-pipeline.groovy) in the Azure jenkins pipelines
- Retain build logs from nightly triggered pipelines from the last 30 builds instead of 10
- Add 'orin-agx-64' node removing 'nuc' and support for related HW tests, bringing hetzci on-par with Azure with this respect. Ref: [PR#453](https://github.com/tiiuae/ghaf-infra/pull/453), Azure [ghaf-hw-test](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/6c4fa3fa02a0ecb26953765ce197d68bb4558d2b/ghaf-hw-test.groovy#L200-L209) and [nightly](https://github.com/tiiuae/ghaf-jenkins-pipeline/blob/6c4fa3fa02a0ecb26953765ce197d68bb4558d2b/ghaf-nightly-pipeline.groovy#L131) pipelines.
- Remove HW-test pipeline support for riscv

### DIFF
--- a/hosts/hetzci/prod/casc/pipelines/ghaf-hw-test.groovy
+++ b/hosts/hetzci/prod/casc/pipelines/ghaf-hw-test.groovy
@@ -50,9 +50,6 @@ def init() {
   } else if(params.IMG_URL.contains("lenovo-x1-")) {
     env.DEVICE_NAME = 'LenovoX1-1'
     env.DEVICE_TAG = 'lenovo-x1'
-  } else if(params.IMG_URL.contains("generic-x86_64-")) {
-    env.DEVICE_NAME = 'NUC1'
-    env.DEVICE_TAG = 'nuc'
   } else if(params.IMG_URL.contains("microchip-icicle-")) {
     env.DEVICE_NAME = 'Polarfire1'
     env.DEVICE_TAG = 'riscv'
--- a/hosts/hetzci/prod/casc/pipelines/ghaf-hw-test.groovy
+++ b/hosts/hetzci/prod/casc/pipelines/ghaf-hw-test.groovy
@@ -50,9 +50,6 @@ def init() {
   } else if(params.IMG_URL.contains("lenovo-x1-")) {
     env.DEVICE_NAME = 'LenovoX1-1'
     env.DEVICE_TAG = 'lenovo-x1'
-  } else if(params.IMG_URL.contains("microchip-icicle-")) {
-    env.DEVICE_NAME = 'Polarfire1'
-    env.DEVICE_TAG = 'riscv'
   } else if(params.IMG_URL.contains("dell-latitude-7330-")) {
     env.DEVICE_NAME = 'Dell7330'
     env.DEVICE_TAG = 'dell-7330'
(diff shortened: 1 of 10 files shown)


hosts/hetzci/vm/casc/pipelines/ghaf-hw-test.groovy
```diff
@@ -50,9 +50,6 @@ def init() {
   } else if(params.IMG_URL.contains("lenovo-x1-")) {
     env.DEVICE_NAME = 'LenovoX1-1'
     env.DEVICE_TAG = 'lenovo-x1'
-  } else if(params.IMG_URL.contains("generic-x86_64-")) {
-    env.DEVICE_NAME = 'NUC1'
-    env.DEVICE_TAG = 'nuc'
   } else if(params.IMG_URL.contains("microchip-icicle-")) {
```
vjuntunen, line 53: Should those riscv stuff to be removed also?

hosts/hetzci/prod/casc/pipelines/ghaf-hw-test.groovy
```diff
@@ -50,9 +50,6 @@ def init() {
   } else if(params.IMG_URL.contains("lenovo-x1-")) {
     env.DEVICE_NAME = 'LenovoX1-1'
     env.DEVICE_TAG = 'lenovo-x1'
-  } else if(params.IMG_URL.contains("generic-x86_64-")) {
-    env.DEVICE_NAME = 'NUC1'
-    env.DEVICE_TAG = 'nuc'
   } else if(params.IMG_URL.contains("microchip-icicle-")) {
```
vjuntunen, line 53: Should those riscv stuff to be removed also?

hosts/hetzci/vm/casc/pipelines/ghaf-hw-test.groovy
henrirosten: True. Removed in https://github.com/tiiuae/ghaf-infra/pull/505/commits/62608739e49bdaa4533877b12f0600061348fe1d

cazfi:
The perf test pipeline was separated from nightly pipeline for a reason. Namely that failures in perf tests (more acceptable) should not show up as failures in nightly pipeline results (less acceptable).

henrirosten:
> The perf test pipeline was separated from nightly pipeline for a reason. Namely that failures in perf tests (more acceptable) should not show up as failures in nightly pipeline results (less acceptable).

You can still see in the nightly test results which test failed, right? To me it seems the point of nightly pipeline is to run more complete, possibly flaky tests and then manually inspect the test results on failure. If we separate the 'perf' to its own pipeline, we would in effect have two nightly pipelines one needs to manually analyze. What's the benefit?

azbeleva:
>  If we separate the 'perf' to its own pipeline, we would in effect have two nightly pipelines one needs to manually analyze. What's the benefit?

We have different types of tests - functional and performance. Nightly pipeline run functional regression tests, and failures there are high priority. If nightly pipeline is not green it makes worry. Performance test failures are more acceptable as was said before, and we do not consider them as failures, more like warning or information. So we don’t want to mix them. This is done for our convenience.

ktusawrk:
Performance tests were separated to a dedicated nightly pipeline because failures are quite common and test results are not 100% deterministic. We need to run perf tests in a dedicated pipeline every night also in the future.

cazfi:
For example, in the prod monitoring we don't care about perf tests at all, but raise a report when nightly pipeline is not green.

cazfi (approved):

PR #573, opened 2025-08-20
hetzci-release: Initial configuration

henrirosten:
Initial configuration for https://ci-release.vedenemo.dev


hosts/monitoring/configuration.nix
```diff
@@ -24,6 +24,7 @@ let
       hetz86-builder
       hetzci-dev
       hetzci-prod
+      hetzci-release
```
joinemm, line 27: hosts/monitoring is deprecated, use hosts/ghaf-monitoring

nix/devshell.nix
```diff
@@ -23,6 +23,12 @@
               --plugins-file "$FLAKE_ROOT"/hosts/azure/jenkins-controller/plugins.txt \
               --output "$FLAKE_ROOT"/hosts/azure/jenkins-controller/plugins.json
           }
+          prefetch-plugins-hetzci-release () {
```
joinemm, line 26: Maybe create one prefetch-plugins function that takes env as its first argument?

henrirosten: True, one way to improve this is in: https://github.com/tiiuae/ghaf-infra/commit/eae2550773c5c3537e4c7e480c0bae01951e00e5

PR #583, opened 2025-08-22
hetzci-prod: Trigger gc about 50GB earlier

cazfi:
Define hetzci-prod min-free and max-free
separately from common.nix, triggering gc when
there's about 70GB free diskspace left.

### DIFF
--- a/hosts/hetzci/prod/configuration.nix
+++ b/hosts/hetzci/prod/configuration.nix
@@ -122,6 +122,8 @@ in
     trusted-public-keys = ghaf-dev.cachix.org-1:S3M8x3no8LFQPBfHw1jl6nmP8A7cVWKntoMKN3IsEQY= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
     substituters = https://ghaf-dev.cachix.org https://cache.nixos.org
     builders-use-substitutes = true
+    min-free =  75000000000
+    max-free = 300000000000
   '';
 
   programs.ssh = {


henrirosten:
> why are we using `nix.extraOptions` in ci-prod configuration instead of `nix.settings`? @henrirosten

@joinemm: 
No real good reason. Hetzci jenkins-controller configuration was originally based on the azure jenkins-controller configuration, which I believe the usage of `nix.extraOptions` dates back to.
I agree we should prefer using `nix.settings` where possible (although, I believe `nix.extraOptions` takes precedence so using `extraOptions` should work correctly too).

@cazfi:
Please somehow state that these values translate to (roughly) 70GB and 280GB and/or maybe indicate it by calculating the values e.g. (70*1024*1024*1024). 
```
    min-free =  75000000000
    max-free = 300000000000
```

hosts/hetzci/dev/configuration.nix
```diff
@@ -122,6 +122,10 @@ in
     trusted-public-keys = ghaf-dev.cachix.org-1:S3M8x3no8LFQPBfHw1jl6nmP8A7cVWKntoMKN3IsEQY= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
     substituters = https://ghaf-dev.cachix.org https://cache.nixos.org
     builders-use-substitutes = true
+    # ~70 GB
+    min-free =  75000000000
+    # ~280 GB
+    max-free = 300000000000
   '';
```
henrirosten, line 129: Proposed change (replace lines 125-129 with):
```
  '';
  nix.settings.min-free = lib.mkForce (70 * 1024 * 1024 * 1024);
  nix.settings.max-free = lib.mkForce (280 * 1024 * 1024 * 1024);
```

hosts/hetzci/prod/configuration.nix
```diff
@@ -122,6 +122,10 @@ in
     trusted-public-keys = ghaf-dev.cachix.org-1:S3M8x3no8LFQPBfHw1jl6nmP8A7cVWKntoMKN3IsEQY= cache.nixos.org-1:6NCHdD59X431o0gWypbMrAURkbJ16ZPMQFGspcDShjY=
     substituters = https://ghaf-dev.cachix.org https://cache.nixos.org
     builders-use-substitutes = true
+    # ~70 GB
+    min-free =  75000000000
+    # ~280 GB
+    max-free = 300000000000
   '';
```
henrirosten, line 129: Proposed change (replace lines 125-129 with):
```
  '';
  nix.settings.min-free = lib.mkForce (70 * 1024 * 1024 * 1024);
  nix.settings.max-free = lib.mkForce (280 * 1024 * 1024 * 1024);
```

henrirosten (commented):
As @joinemm hinted, let's also prepare a fix to move all the other `nix.extraOptions` to `nix.settings` also (not necessarily in this PR).

henrirosten (approved):

PR #590, opened 2025-08-28
hetzci: Init release pipeline

henrirosten:
- Add initial release pipeline for `vm`, `dev` and `release` environments
- Remove ghaf-cachix-pin pipeline, since the pinning now happens part of release pipeline. Cachix pin is now done on the [ghaf-dev](https://app.cachix.org/organization/tiiuae/cache/ghaf-dev#pins) cache; eventually, we want to have a separate release cachix cache, but the release cache and its pinning will be implemented in follow-up PRs.

Deployed and tested here: https://ci-release.vedenemo.dev/job/ghaf-release/2/


joinemm (approved):

henrirosten:
One more push: it's better to run the OTA build/pin _after_ image builds, not before. 

When OTA build happens after image builds, overall pipeline execution time reduces because when built in this order, most OTA target package dependencies are already cached in local store. 

In addition, also added the release pipeline to `dev` environment.

joinemm (approved):

PR #594, opened 2025-08-29
Refactor hetzci into reusable modules

joinemm:
Split up hetzci into modules based on what they implement. Each environment (dev, prod, ...) just imports the implementations they want. Any additional host specific configuration can be added into the configuration.nix, or an option can be added to one of the modules.

For now each environment still retains its own jenkins casc, pipelines, plugins and secrets.


henrirosten (approved):
This looks good, thanks @joinemm.

While, after this PR, it will be easier to accidentally introduce a change to an environment it was not intended to, this refactor does simplify and clean the config.

On making changes to now 'common' parts, we just have to be careful that the change is really meant to all environments.

joinemm:
> On making changes to now 'common' parts, we just have to be careful that the change is really meant to all environments.

@henrirosten I'm hoping this would get caught at least in code review

PR #597, opened 2025-09-01
Add UEFI SecureBoot Signing

alextserepov:
- offline keys created for each build

### DIFF
--- a/hosts/hetzci/prod/casc/pipelines/modules/utils.groovy
+++ b/hosts/hetzci/prod/casc/pipelines/modules/utils.groovy
@@ -14,6 +14,19 @@ def append_to_build_description(String text) {
   }
 }
 
+def runUEFISign(String diskPath, String target) {
+  def outdir = "${target}-signed"
+  sh """
+    mkdir -p "${outdir}/keys"
+    uefikeygen
+    test -f keys/db.crt -a -f keys/db.key
+    cp -v keys/*.der "${outdir}/keys/"
+ 
+    uefisign keys/db.crt keys/db.key "${diskPath}" "${outdir}"
+  """
+  return run_cmd("nix store add ${outdir}")
+}
+
 def create_pipeline(List<Map> targets) {
   def pipeline = [:]
   def stamp = run_cmd('date +"%Y%m%d_%H%M%S%3N"')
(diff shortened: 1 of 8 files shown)


henrirosten (commented):
One general note in addition to the more specific ones below: have you considered adding this feature to the target (Ghaf) repo? That is, Ghaf would have an output, say, `packages.x86_64-linux.lenovo-x1-carbon-gen11-signed` that would have the recipe to generate such signed images?

Not sure if we need to, but how about OTA - can we have uefi signed OTA images with the current architecture?

hosts/hetzci/vm/casc/pipelines/modules/utils.groovy
```diff
@@ -14,6 +14,28 @@ def append_to_build_description(String text) {
...
+    nix run --accept-flake-config --option builders '' --option max-jobs 1 \
+      github:tiiuae/ci-yubi#uefikeygen --
+
+    nix run --accept-flake-config --option builders '' --option max-jobs 1 \
+      github:tiiuae/ci-yubi#uefisign -- \
+      keys/db.crt keys/db.key "${diskPath}" "signed"
+
+    test -f keys/db.crt -a -f keys/db.key
+    cp -v keys/*.der "${outDir}/keys/"
+    cp -v signed/* "${outDir}"
```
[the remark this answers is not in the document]
joinemm: I tested this and jenkins purged the artifacts immediately

joinemm:
I force pushed my changes into this PR. scripts are now sourced from ci-yubi input instead of using nix run.

hosts/hetzci/vm/casc/pipelines/modules/utils.groovy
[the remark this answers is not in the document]
joinemm: now the signed image is a nix store symlink also

hosts/hetzci/prod/casc/pipelines/modules/utils.groovy
```diff
@@ -77,6 +90,16 @@ def create_pipeline(List<Map> targets) {
           }
         }
       }
+      if (it.get('uefisign', false)) {
+        stage("UEFI Sign ${shortname}") {
+           def disk_path  = run_cmd("find -L ${it.target} -type f -name 'disk1.raw.zst' -print -quit")
```
henrirosten, line 95: This will only work with small subset of Ghaf outputs that produce `disk1.raw.zst`, but I guess that's fine for now?

henrirosten (changes requested):
In addition to the below comment, I notice one other issue that I think we should fix before merging this: the added 'UEFI Sign' step runs some minutes, maybe around 2-5 minutes typically. If the service `jenkins-purge-artifacts.service` happens to run during this time, it will remove the artifacts directory and the generated provenance file because there's no symlinks in the directory at that point.

This is not a bug you introduced, but extra delay caused by the added 'UEFI sign' step makes this issue more apparent.

hosts/hetzci/prod/casc/pipelines/modules/utils.groovy
joinemm: yes because afaik the signing script itself expects only that format

henrirosten: @alextserepov: Something to improve in future revisions?

joinemm (approved):

PR #605, opened 2025-09-08
Remove ficolo servers other than builders

cazfi:


hosts/ficolo-common.nix
```diff
@@ -1,22 +1,9 @@
...
-{
-  self,
-  lib,
-  machines,
-  ...
-}:
-{
-  imports = [ self.nixosModules.service-monitoring ];
-
+_: {
```
henrirosten, line 4: `_: ` in the beginning of line 4 is unnecessary leftover and should be removed

hosts/builders/ficolo.nix
```diff
@@ -40,11 +40,6 @@
     };
   };
 
-  services.monitoring = {
-    metrics.enable = true;
-    logs.enable = true;
-  };
-
```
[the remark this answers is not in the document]
cazfi: It was a collapsing dependency chain that lead to this. IIRC it started from the removed IP address definition of the ficolo-monitoring server.

hosts/ficolo-common.nix
cazfi: Fixed

henrirosten (approved):

PR #619, opened 2025-09-16
Initial version of cachix-push.nix

henrirosten:
- Add `cachix-push.nix` nix module which implements a systemd service that is functionally-wise similar to [cachix-watch-store](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/system/cachix-watch-store.nix), but allows filtering the store paths pushed to cachix. This service can be used to replace the [cachix-watch-store](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/system/cachix-watch-store.nix) service (which we currently use e.g. in [hetzarm](https://github.com/tiiuae/ghaf-infra/blob/main/hosts/builders/hetzarm/configuration.nix#L44-L49) and [hetz86-1](https://github.com/tiiuae/ghaf-infra/blob/main/hosts/builders/hetz86-1/configuration.nix#L46-L51) builders).
- Start using `cachix-push.nix` in the release builders `hetz86-rel-1` and `hetzarm-rel-1`, pushing the build results to [ghaf-release](https://app.cachix.org/organization/tiiuae) cachix cache
- Remove OTA pins "nvidia-jetson-orin-nx-debug-from-x86_64" and "nvidia-jetson-orin-nx-debug" from the release pipeline that were added only to keep the release cache warm for cross-compiled builds that otherwise would take ages. This is now unnecessary, since we push all build results to ghaf-release cachix cache from the release builders.

A follow-up PR will disable [cachix-watch-store](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/system/cachix-watch-store.nix) service in [hetzarm](https://github.com/tiiuae/ghaf-infra/blob/main/hosts/builders/hetzarm/configuration.nix#L44-L49) and [hetz86-1](https://github.com/tiiuae/ghaf-infra/blob/main/hosts/builders/hetz86-1/configuration.nix#L46-L51) builders replacing it with `cachix-push.nix`.

Note about the cachix push [filtering](https://github.com/tiiuae/ghaf-infra/blob/78a79748b5c208f2812f73b826f14bd57f69a34f/hosts/builders/cachix-push.sh#L80-L81): we skip pushing the images, but otherwise push the full closure. This significantly reduces the cachix storage requirement. As an example, the current storage used on [ghaf-release](https://app.cachix.org/organization/tiiuae) cachix cache is  around 30GB - this is the amount of storage required from one full Ghaf release build (7 image targets + doc) after the changes from this PR.

This change has been tested in https://ci-release.vedenemo.dev/.

### DIFF
--- a/hosts/builders/cachix-push.sh
+++ b/hosts/builders/cachix-push.sh
@@ -1,0 +1,87 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2022-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-FileCopyrightText: 2018 GitHub, Inc. and contributors
+# SPDX-License-Identifier: Apache-2.0
+
+set -e # exit immediately if a command fails
+set -E # exit immediately if a command fails (subshells)
+set -u # treat unset variables as an error and exit
+
+# Temporary workdir for the script
+TMPDIR="$(mktemp -d --suffix .cachix-push)"
+
+# Expected arguments and their defaults if not passed in environment variables
+CACHIX_AUTH_TOKEN_FILE="${CACHIX_AUTH_TOKEN_FILE:=/dev/null}"
+CACHIX_CACHE_NAME="${CACHIX_CACHE_NAME:=ghaf-dev}"
+
+# Lists all nix store paths potentially pushed to cachix
+list_nix_store_paths () {
+    out=$1
+    echo -n "" >"$out"
+    # Following is (mostly) copied from:
+    # https://github.com/cachix/cachix-action/blob/ee79d/dist/list-nix-store.sh
+    for file in /nix/store/*; do
+        case "$file" in
+        *.drv)
+            # Avoid .drv as they are not generally useful
+            continue
+            ;;
+        *.drv.chroot)
+            # Avoid .drv.chroot as they are not generally useful
+            continue
+            ;;
+        *.check)
+            # Skip .check file produced by --keep-failed
+            continue
+            ;;
+        *.lock)
+            # Skip .lock files
+            continue
+            ;;
+        *.links)
+            # Skip .links file populated by nix-store --optimize
+            continue
+            ;;
+        *)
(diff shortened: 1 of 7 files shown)


brianmcgillion:
very nice!

hosts/builders/cachix-push.sh
```diff
@@ -0,0 +1,87 @@
...
+    list_nix_store_paths "$TMPDIR/snap"
+    # Diff the ref and current snapshot to find new store paths since the last
+    # diff. Use line group formats to output only new and changed lines.
+    # For detailed description, see diff option --GTYPE-group-formats manual:
+    diff \
+        --unchanged-group-format='' --changed-group-format='%>' \
+        --old-group-format='' --new-group-format='%>' \
+        "$TMPDIR/ref" "$TMPDIR/snap" >"$TMPDIR/new" || true
+    # Filter out paths that match the following regexp
+    filter='(\.img$|\.raw$|\.zst$|\.iso$|-disko-image)'
```
cazfi, line 81: Do we really want to filter *all* .zst compressed files, or should we have filters like .iso.zst?

cazfi (changes requested):

hosts/builders/cachix-push.sh
henrirosten: That's a good point. I'll do some more testing still and adjust the filter accordingly.

henrirosten: Fixed in https://github.com/tiiuae/ghaf-infra/pull/619/commits/e3ba0d54a82368f04bb80d2d793061cb3a6d5ec3.

henrirosten:
Pushed a version that:
- Improves `cachix-push` filter and also skips cachix push if the store path is a (symlink to) directory and any directory contents matches the filter
- Improves error handling
- Automatically restarts the `cachix-push` service if the service stops for one reason or another

These changes have been tested in https://ci-release.vedenemo.dev/.

cazfi (approved):

PR #620, opened 2025-09-17
Add UEFI SB Signing to Darter target

alextserepov:


henrirosten (approved):
Assuming this has been somehow tested.

PR #622, opened 2025-09-18
Remove older profile generations

cazfi:
Keep just 5 latest profile generations. Remove older ones nightly.

Tested on ficolo hosts.

### DIFF
--- a/hosts/common.nix
+++ b/hosts/common.nix
@@ -110,6 +110,15 @@ in
   # Shell
   programs.bash.completion.enable = true;
 
+  systemd.services.trim-profiles = {
+    description = "Delete older profiles";
+    serviceConfig.Type = "oneshot";
+    script = ''
+      ${pkgs.nix}/bin/nix-env --delete-generations +3 --profile /nix/var/nix/profiles/system
+    '';
+    startAt = "03:00";
+  };
+
   # https://nixos.wiki/wiki/FAQ/When_do_I_update_stateVersion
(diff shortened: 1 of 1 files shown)


hosts/common.nix
```diff
@@ -110,6 +110,15 @@ in
   # Shell
   programs.bash.completion.enable = true;
 
+  systemd.services.trim_profiles = {
```
henrirosten, line 113: Suggested change (line 113):
```
  systemd.services.trim-profiles = {
```
That is, rename the service to 'trim-profiles' to follow the naming convention used with other systemd services.

hosts/common.nix
```diff
@@ -110,6 +110,15 @@ in
...
+    description = "Delete older profiles";
+    serviceConfig.Type = "oneshot";
+    script = ''
+      /run/current-system/sw/bin/nix-env --delete-generations +5 --profile /nix/var/nix/profiles/system
```
henrirosten, line 117: Suggested change (line 117):

```
      ${pkgs.nix}/bin/nix-env --delete-generations +3 --profile /nix/var/nix/profiles/system
```

That is:
- Use `${pkgs.nix}/bin/nix-env` instead of `/run/current-system/sw/bin/nix-env`
- Keep only the last 3 generations which to me seems enough for our use

henrirosten (approved):

joinemm (approved):

PR #632, opened 2025-09-24
Trial provenance fix

henrirosten:
Trial a fix for the provenance generation issue, an example [here](https://ci-prod.vedenemo.dev/job/ghaf-nightly/117/pipeline-overview/?selected-node=593).

This PR includes two fixes:
- Use build out-path (instead of flake reference) as the provenance target. This should ensure the target is not re-evaluated, to make sure the provenance target is the same as the build output.
- Include a fix to provenance tool from the [sbomnix/fix-provenance](https://github.com/tiiuae/sbomnix/compare/main...fix-provenance) branch, which ensures the provenance subject is force-realised making sure the provenance target is always available in local nix store.

The issue is hard to reproduce, which is why I propose we merge and deploy this change, then monitor if we still see the issue. We could conclude the issue is fixed if it doesn't emerge after these changes have been deployed and running in prod for two weeks. If the fix is successful, we would merge the [sbomnix/fix-provenance](https://github.com/tiiuae/sbomnix/compare/main...fix-provenance) branch and move the ghaf-infra sbomnix input back to sbomnix main.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -231,11 +231,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1755960406,
-        "narHash": "sha256-RF7j6C1TmSTK9tYWO6CdEMtg6XZaUKcvZwOCD2SICZs=",
+        "lastModified": 1757239681,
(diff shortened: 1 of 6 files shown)


joinemm (approved):
let's see if it works

PR #633, opened 2025-09-25
Fix downloading of agent.jar

joinemm:
The wget command never worked but we didn't notice it because the agent.jar was always there

### DIFF
--- a/hosts/testagent/agent.nix
+++ b/hosts/testagent/agent.nix
@@ -60,10 +60,8 @@ let
         ExecStart = toString (
           pkgs.writeShellScript "jenkins-connect.sh" # sh
             ''
-              set -x
-
               if [[ -z "$CONTROLLER" ]]; then
-                echo "ERROR: Variable CONTROLLER not configured in $(pwd)/jenkins.env"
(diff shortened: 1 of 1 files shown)


hosts/testagent/agent.nix
```diff
@@ -123,14 +123,15 @@ in
           serviceConfig = {
             Type = "oneshot";
             User = "jenkins";
+            EnvironmentFile = "/var/lib/jenkins/jenkins.env";
             RemainAfterExit = "yes";
             WorkingDirectory = "/var/lib/jenkins";
             ExecStart = toString (
               pkgs.writeShellScript "start-agents.sh" # sh
                 ''
                   if [[ ! -f agent.jar ]]; then
```
cazfi, line 132: This means that the new version of the .jar is not downloaded after the Jenkins version on the controller has been updated. Maybe we should download it always.

cazfi (changes requested):

hosts/testagent/agent.nix
joinemm: hmm that is true

PR #645, opened 2025-10-03
Retry provenance on failure

henrirosten:
There's something strange that occasionally fails in the provenance creation. I'd like to see if running the provenance again on failure makes any difference.


cazfi (approved):

ktusawrk (approved):
Bit of a kludge but let's try if it helps

PR #649, opened 2025-10-06
Add SoftHSM capabilities to nethsm-gateway

joinemm:
Added softhsm to nethsm-gateway and instructions how to use it. We can interact with the softhsm by using a softhsm module with pkcs11, and openssl can interact with pkcs11 using pkcs11-provider. This is a new openssl 3.X provider.

I was able to remove the usage of legacy openssl features throughout the whole signing process by using `systemd-sbsign` instead of the old `sbsign` from sbsigntools. `systemd-sbsign` can use openssl 3.X providers to access key objects straight from the softhsm.

### DIFF
--- a/users/jrautiola.nix
+++ b/users/jrautiola.nix
@@ -13,6 +13,7 @@
       extraGroups = [
         "wheel"
         "networkmanager"
+        "softhsm"
       ];
     };
   };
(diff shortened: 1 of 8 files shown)


alextserepov (dismissed):

users/jrautiola.nix
```diff
@@ -13,6 +13,7 @@
       extraGroups = [
         "wheel"
         "networkmanager"
+        "softhsm"
```
cazfi, line 16: You should give this group to some more people, Alex at least.

cazfi (approved):

henrirosten (approved):

PR #920, opened 2026-03-02
flake.lock: Update flake inputs and Jenkins plugins

ssrcdevops:
This PR updates all flake inputs in `flake.lock` and refreshes Jenkins plugins manifests:
- `hosts/hetzci/plugins.json`
- `hosts/uae/azureci/plugins.json`

PR #997, opened 2026-04-09
March -26 security update

cazfi:
Update inputs and jenkins plugins to latest versions.
Changes have gone through inputs update / security update testing process.
To make it really work, a bit different changes were required compared to automatically generated #920

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -31,11 +31,11 @@
         "sbsigntools": "sbsigntools"
       },
       "locked": {
-        "lastModified": 1769770935,
-        "narHash": "sha256-m0LUroArijpBACfc+uCRGNZ1ScXjeG3RCxtqrKPkyRA=",
+        "lastModified": 1772562307,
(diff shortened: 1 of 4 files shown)


henrirosten (changes requested):
Please add a PR description covering:

  - what changed and why
  - how this was tested
  - why this needed a separate PR instead of [PR#920](https://github.com/tiiuae/ghaf-infra/pull/920)

I don’t see issues in the patch itself, but I belive that context should be in the PR for review and future traceability.

cazfi:
Because of SSO I can not push directly to tiiuae/ghaf-infra repository, to update #920 with the required changes; adding kexec commit before the inputs update & having inputs update to a newer nikpkgs revision than with the latest automatically created commit, to include fix to the critical security vulnerability.

henrirosten (approved):

joinemm (approved):

PR #1003, opened 2026-04-14
Unify host inventory

henrirosten:
Collapses the three parallel host sources (`hosts/machines.nix`, `hostModules` in `hosts/default.nix`, and the per-arch node sets in `nix/deployments.nix`) into a single canonical inventory in `hosts/machines.nix`. Each entry carries `module`, `system`, optional `machine` (deploy metadata), and optional `kind = "vm"` for outliers.

Consumers derive their view from the inventory:

- `hosts/default.nix` auto-generates `flake.nixosConfigurations` and `flake.nixosModules`, and injects `nixpkgs.hostPlatform` from `system` (removes ~30 duplicated per-host lines).
- `nix/deployments.nix` derives per-arch node sets by filtering on `system`.
- A new `flake.lib.hostsBySystem` drives the build filter in `.github/workflows/test-ghaf-infra.yml`, replacing the `hetzarm`-substring regex so aarch64 hosts no longer need that name prefix.

Adding a host is now one inventory entry plus the configuration module. `docs/adding-a-host.md` updated accordingly (10 steps reduced to 8).

Also adds `workflow_dispatch` to `test-ghaf-infra.yml` so future workflow changes can be exercised on a branch before merge via`gh workflow run test-ghaf-infra.yml --ref <branch>` (this PR's own workflow changes cannot be tested that way because the trigger was only added in this branch).

#### How this has been tested
  - [x] `nix flake check --no-build --option allow-import-from-derivation false`
  - [x] `nix fmt`
  - [x] `nix eval .#nixosConfigurations --apply builtins.attrNames`: set of configs is identical pre/post (32 entries).
  - [x] Simulated both the old and the new CI build filters against the  target list produced by the extraction logic in `scripts/nix-fast-build.sh`. Target sets on `x86_64-linux` and `aarch64-linux` are identical to the old behavior.
  - [x] CI build (manually triggered) on x86_64 builder
  - [x] CI build (manually triggered)  on aarch64 builder
  - [x] Spot-deploy one host (hetzci-dbg) with `deploy-rs` to confirm the inventory-driven node metadata still resolves correctly.

### DIFF
--- a/hosts/default.nix
+++ b/hosts/default.nix
@@ -14,6 +14,15 @@ let
   );
   isAutoConfiguredHost = _: host: (host.kind or "host") != "vm";
 
+  # Per-system grouping of nixosConfiguration names. Driven by the inventory's
+  # `system` field. Seeded with `hetzci-vm-no-host-nix-store`, which is an
+  # x86_64-linux VM variant exposed in `flake.nixosConfigurations` without its
+  # own inventory entry.
+  hostsBySystem = lib.foldlAttrs (
+    acc: name: host:
+    acc // { ${host.system} = (acc.${host.system} or [ ]) ++ [ name ]; }
+  ) { x86_64-linux = [ "hetzci-vm-no-host-nix-store" ]; } hostInventory;
+
   # make self and inputs available in nixos modules
   specialArgs = {
     inherit self inputs machines;
@@ -86,9 +95,8 @@ in
   }
   // nixosModulesFromHosts;
 
-  # Expose as flake.lib.mkNixOS.
   flake.lib = {
-    inherit mkNixOS;
+    inherit mkNixOS hostsBySystem;
   };
 
   flake.nixosConfigurations = nixosConfigurationsFromHosts // {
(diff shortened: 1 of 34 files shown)


ktusawrk (dismissed):

hosts/default.nix
```diff
@@ -64,48 +80,12 @@ let
...
-    uae-azureci-az86-1 = ./uae/azureci/builders/az86-1/configuration.nix;
-    uae-testagent-prod = ./uae/testagent/prod/configuration.nix;
-    uae-azureci-hetzarm-1 = ./uae/azureci/builders/hetzarm-1/configuration.nix;
-  };
-
   nixosModulesFromHosts = lib.mapAttrs' (
-    name: path: lib.nameValuePair "nixos-${name}" path
-  ) hostModules;
+    name: host: lib.nameValuePair "nixos-${name}" host.module
+  ) hostInventory;
```
joinemm: do we still need the hosts as nixos modules? iirc it was only used for extending config for azure VMs.

henrirosten: True, that's not needed. Fixed in: 09aa7c1caf78dde6160cf83fcce036fa290d55b4

joinemm (approved):

PR #1007, opened 2026-04-16
jenkins: migrate utils.groovy to shared library

henrirosten:
**Summary**:
  - Migrate utils.groovy from ad-hoc load() to a proper Jenkins shared library (@Library('ghafInfra')) 
  - Register the library via a CasC init script backed by a Nix-provisioned synthetic Git repo under /etc/jenkins/pipeline-library
  - Update all pipeline scripts to use the @Library annotation instead of the MODULES load pattern
 
**Motivation**:
The previous approach used load "/etc/jenkins/pipelines/modules/utils.groovy" which only works on the controller node. By switching to a Jenkins shared library, utils.groovy is resolved through the standard @Library mechanism, which works on any node. This enables using utils.groovy (and possible other future shared libraries) from hw-test agents: a separate PR will follow to refactor hw-test pipelines so they also use the shared library (edit: [already available here](https://github.com/tiiuae/ghaf-infra/pull/1012)).

**Security / Risk**:
To let Jenkins register the Nix-provisioned shared library from `file:///etc/jenkins/pipeline-library`, this change enables `-Dhudson.plugins.git.GitSCM.ALLOW_LOCAL_CHECKOUT=true` on the controller.

This is an intentional controller-wide policy relaxation compared to the previous design, which loaded one fixed local Groovy file directly. In our deployment, this surface remains admin-only: authenticated users have read access, while only the `tiiuae:devenv-fi` admin group can administer Jenkins and define or modify SCM-backed jobs/libraries. This is an acceptable tradeoff for the shared-library design, but it is a real security-boundary change and worth calling out explicitly.

**How this was tested**
See: https://ci-dbg.vedenemo.dev/job/ghaf-manual/

### DIFF
--- a/hosts/hetzci/casc/common.yaml
+++ b/hosts/hetzci/casc/common.yaml
@@ -75,6 +75,30 @@ groovy:
       import java.util.logging.Level
       import java.util.logging.Logger
       Logger.getLogger("hudson.security.csrf.CrumbFilter").setLevel(Level.SEVERE)
+  # Register the deployment-bundled shared library explicitly. Pipelines use
+  # @Library('ghafInfra') _, so Jenkins must expose /etc/jenkins/pipeline-library
+  # through the shared-library subsystem rather than via ad hoc load() calls.
+  - script: |
+      import jenkins.model.Jenkins
+      import jenkins.plugins.git.GitSCMSource
+      import org.jenkinsci.plugins.workflow.libs.GlobalUntrustedLibraries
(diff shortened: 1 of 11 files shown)


henrirosten:
Rebased now that https://github.com/tiiuae/ghaf-infra/pull/1009 landed.

Re-tested in ci-dbg, e.g.:
- https://ci-dbg.vedenemo.dev/job/ghaf-manual/14/
- https://ci-dbg.vedenemo.dev/job/ghaf-manual/15/

ktusawrk (approved):

PR #1008, opened 2026-04-16
slsa: trust provenance from hetzci-dbg

henrirosten:
The provenance trust policy accepts the dev, prod and release controllers, but it omitted the debug controller.

Add ci-dbg.vedenemo.dev to the allowed builder IDs so test agents accept provenance generated by hetzci-dbg.


ktusawrk (approved):
Test automation must be available also for hetzci-dbg, so this change is needed.

joinemm (approved):

PR #1021, opened 2026-04-21
Use S3 backend and configure retention for zot registry

joinemm:
Artifacts are now stored in hetzner object storage. Retention rules are quite conservative but let's see what the disk usage lands at and tweak them as needed.

### DIFF
--- a/hosts/ghaf-registry/configuration.nix
+++ b/hosts/ghaf-registry/configuration.nix
@@ -10,13 +10,62 @@
 }:
 let
   inherit (self.packages.${pkgs.stdenv.hostPlatform.system}) zot;
+
   zotDataDir = "/var/lib/zot";
   zotPort = 5000;
+
   zotConfig = {
     storage = {
       rootDirectory = zotDataDir;
+
+      storageDriver = {
+        name = "s3";
+        bucket = "oci-artifacts";
+        region = "hel1";
+        forcepathstyle = true;
+        regionendpoint = "https://hel1.your-objectstorage.com";
+        chunksize = toString (32 * 1024 * 1024);
+      };
+
+      dedupe = false;
       gc = true;
-      gcDelay = "1h";
+      gcInterval = "1h";
+      retention = {
+        delay = "24h";
+        policies = [
+          {
+            repositories = [ "ghaf/**" ];
+            deleteReferrers = true;
+            deleteUntagged = true;
+            keepTags = [
+              {
+                patterns = [ "release-.*" ];
+                # Keep at least three builds in each repository
+                mostRecentlyPushedCount = 3;
+                # Keep everything for at least a week
+                pushedWithin = "168h";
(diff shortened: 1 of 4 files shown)


ktusawrk:
What is the data migration plan from /var/lib/zot to s3://oci-artifacts? Do we even need to migrate any data at this point?

joinemm:
No need to migrate anything at this point, s3 was initialized from empty and disk was wiped

joinemm:
added proxy config that seems to fix the intermittent pull failures

hosts/ghaf-registry/configuration.nix
```diff
@@ -10,33 +10,80 @@
...
+            deleteUntagged = true;
+            keepTags = [
+              {
+                patterns = [
+                  "release-.*"
+                  "prod-.*"
+                  "dev-.*"
+                ];
+                # Keep at least three builds in each repository, meaning 3 builds per pipeline
+                mostRecentlyPushedCount = 3;
```
henrirosten: Did you intend mostRecentlyPushedCount = 3 to be shared across release-*, prod-*, and dev-* here? In zot, the count is applied per keepTags entry, not per pattern, so those prefixes share one top-3 budget. In practice, frequent dev pushes could consume those 3 slots and allow older prod or release tags in the same repository to be collected sooner than expected.

joinemm: ah I didn't realize it would be per keepTags entry. I'll separate those to one env per entry.

hosts/ghaf-registry/configuration.nix
```diff
@@ -10,33 +10,86 @@
...
   zotPort = 5000;
+
   zotConfig = {
     storage = {
       rootDirectory = zotDataDir;
+
+      storageDriver = {
+        name = "s3";
+        bucket = "oci-artifacts";
+        region = "eu-central";
```
henrirosten: I _believe_ this should be `hel1`, not `eu-central`. Hetzner Object Storage uses the location code as the S3 signing region, and with `regionendpoint = "https://hel1.your-objectstorage.com"` this mismatch can cause auth/signature failures. 

Hetzner’s docs use the location code (`hel1`/`fsn1`/`nbg1`) as the S3 region, with matching location-specific endpoints like `hel1.your-objectstorage.com`. `eu-central` is the Hetzner network zone, not the S3 region:
  - https://docs.hetzner.com/storage/object-storage/getting-started/using-curl/
  - https://docs.hetzner.com/storage/object-storage/getting-started/using-s3-api-tools/
  - https://docs.hetzner.com/storage/object-storage/overview/

Suggested change: `region = "hel1";`

joinemm: that sounds right, but it's strange that it still works with the current setup.

henrirosten (approved):

PR #1025, opened 2026-04-23
Add FLASHED variable for robot

leivos-unikie:
If this variable is true robot can expect boot from fresh ghaf image.

See: https://github.com/tiiuae/ci-test-automation/pull/777

### DIFF
--- a/hosts/hetzci/pipelines/ghaf-hw-test-manual.groovy
+++ b/hosts/hetzci/pipelines/ghaf-hw-test-manual.groovy
@@ -160,6 +160,7 @@ def ghaf_robot_test(String tags) {
           -v DEVICE_TYPE:$DEVICE_TAG \
           -v BUILD_ID:${BUILD_NUMBER} \
           -v COMMIT_HASH:NONE \
+          -v FLASHED:${FLASHED} \
           -i $INCLUDE_TEST_TAGS .
       '''
       currentBuild.description = "${currentBuild.description}<br>✅ ${tags}"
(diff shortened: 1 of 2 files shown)


hosts/hetzci/pipelines/ghaf-hw-test.groovy
```diff
@@ -276,6 +277,7 @@ pipeline {
           steps {
             script {
               env.TEST_CONFIG_DIR = 'Robot-Framework/config'
+              env.FLASHED = '${False}'
```
henrirosten: Passing `${False}` / `${True}` here will break the Robot-side consumer added in tiiuae/ci-test-automation#777

`ghaf-robot` passes CLI variables as strings, and the new Robot code uses `IF ${FLASHED}`. With the values from this patch, Robot receives the literal strings `${False}` / `${True}`, so the condition expands to `IF ${False}` / `IF ${True}` and fails with `SyntaxError` instead of behaving like a boolean.

Please pass plain `False` / `True` instead.

hosts/hetzci/pipelines/ghaf-hw-test-manual.groovy
```diff
@@ -258,6 +259,7 @@ pipeline {
           steps {
             script {
               env.TEST_CONFIG_DIR = 'Robot-Framework/config'
+              env.FLASHED = '${False}'
```
henrirosten: Passing `'${False}'` / `'${True}'` here will break the Robot-side consumer added in tiiuae/ci-test-automation#777

`ghaf-robot` passes CLI variables as strings, and the new Robot code uses `IF ${FLASHED}`. With the values from this patch, Robot receives the literal strings `'${False}'` / `'${True}'`, so the condition expands to `IF '${False}'` / `IF '${True}'` and fails with `SyntaxError` instead of behaving like a boolean.

Please pass plain `'False'` / `'True'` instead.

henrirosten (approved):

PR #1027, opened 2026-04-25
Robots: Switch to systemd-boot bootloader

cazfi:
Booting with grub didn't work, but they were fallbacking to old systemd-boot bootloader anyway. Now we explicitly use systemd-boot, and more importantly keep systemd-boot setup updated with new kernels and bootloader versions.

Note that manual onetime actions might be needed on the host with old grub boot loader installation when migrating to a revision with systemd-boot loader.
Tested on hetz86-builder, and there old /boot/EFI/BOOT/BOOTX64.EFI had to be temporarily renamed for the migratory deployment to pass.

### DIFF
--- a/hosts/hetzner-robot.nix
+++ b/hosts/hetzner-robot.nix
@@ -34,9 +34,8 @@
     networkConfig.DHCP = "ipv4";
   };
 
-  boot.loader.grub = {
-    efiSupport = true;
-    efiInstallAsRemovable = true;
+  boot.loader = {
+    systemd-boot.enable = true;
   };
 
   environment.systemPackages = with pkgs; [
--- a/hosts/hetzner-robot.nix
+++ b/hosts/hetzner-robot.nix
@@ -34,9 +34,9 @@
     networkConfig.DHCP = "ipv4";
   };
 
-  boot.loader.grub = {
-    efiSupport = true;
-    efiInstallAsRemovable = true;
+  boot.loader = {
+    systemd-boot.enable = true;
+    systemd-boot.configurationLimit = 5;
   };
 
   environment.systemPackages = with pkgs; [


hosts/hetzner-robot.nix
```diff
@@ -34,9 +34,8 @@
     networkConfig.DHCP = "ipv4";
   };
 
-  boot.loader.grub = {
-    efiSupport = true;
-    efiInstallAsRemovable = true;
+  boot.loader = {
+    systemd-boot.enable = true;
```
henrirosten, line 38: Please set a bounded systemd-boot generation limit. These hosts use a 512 MiB ESP mounted at /boot, and systemd-boot copies kernel/initrd files there. Live inspection showed roughly 24 MiB per x86_64 generation and about 70 MiB for the observed aarch64 generation on hetzarm, with stale GRUB files also still present. To keep rollback depth while avoiding /boot filling over repeated deploys, I’d suggest:

```
  boot.loader = {
    systemd-boot = {
      enable = true;
      configurationLimit = 5;
    };
  };
```

henrirosten, line 38: The PR description notes that “manual onetime actions might be needed on the host with old grub boot loader installation,” and live inspection confirms that is true for both `hetz86-rel-2` and `uae-azureci-hetzarm-1`. They both currently report `systemd-boot not installed in ESP` from `bootctl --esp-path=/boot status`, have no `/boot/loader/entries`, and are booting via GRUB-style `/boot/kernels/...` paths.

Both need migration, otherwise deploy of this shared module may fail.

cazfi: Done

cazfi: Yes

henrirosten (approved):

PR #1042, opened 2026-05-02
zot metrics monitoring

cazfi:
Add zot metrics monitoring, and a new grafana dashboard for them.

### DIFF
--- a/hosts/ghaf-registry/configuration.nix
+++ b/hosts/ghaf-registry/configuration.nix
@@ -75,7 +75,7 @@ let
       externalUrl = "https://registry.vedenemo.dev";
 
       auth = {
-        htpasswd.path = config.sops.secrets.zot-htpasswd.path;
+        htpasswd.path = config.sops.templates."zot-htpasswd".path;
 
         openid.providers.oidc = {
           name = "Vedenemo Auth";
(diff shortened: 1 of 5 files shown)


hosts/ghaf-registry/configuration.nix
```diff
@@ -155,6 +156,14 @@ in
...
       zot-s3-credentials.owner = "zot";
       zot-htpasswd.owner = "zot";
+      zot-metrics-htpasswd-entry.owner = "zot";
+    };
+    templates."zot-htpasswd" = {
+      owner = "zot";
+      content = ''
+        ${config.sops.placeholder.zot-htpasswd}
+        ${config.sops.placeholder."zot-metrics-htpasswd-entry"}
+      '';
```
joinemm: I don't see much point in this template that combines two secrets into one. both htpasswd entries could just be under the same `zot-htpasswd` secret on separate lines.

joinemm (changes requested):

hosts/ghaf-registry/configuration.nix
cazfi: Fixed

joinemm (approved):

PR #1046, opened 2026-05-07
Implement signing redundancy router

joinemm:
Adds a PKCS11 signing redundancy router for the CI signing pipeline. Signing now selects the first healthy signing proxy/token combination for each requested key object instead of assuming one fixed token and socket. A node is considered "healthy" if p11tool can find the requested key object. The socket and key that were used are then recorded into the manifest.

The router prefers signing nodes in the following order:
- Tampere NetHSM
- Masdar NetHSM
- Tampere YubiHSM
- Masdar YubiHSM

The tunnel into Masdar is not functional at the moment so that part was left untested, but in theory it should work.

Note that as of now, the public keys used in the pipelines are hardcoded still to be the Tampere YubiHSM (coming from ghaf-infra-pki).

Tested at https://ci-dbg.vedenemo.dev/job/ghaf-manual/5/

### DIFF
--- a/scripts/select-pkcs11-node.sh
+++ b/scripts/select-pkcs11-node.sh
@@ -1,0 +1,127 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: 2022-2025 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# Detect whether this file is sourced or executed
+_sourced=0
+if [[ ${BASH_SOURCE[0]} != "$0" ]]; then
+  _sourced=1
+fi
+
+# Save caller's shell option state before changing it
+_old_shell_opts="$(set +o)"
+_old_shopt_opts="$(shopt -p)"
+
+cleanup_shell_state() {
+  local rc=$?
+  eval "$_old_shell_opts"
+  eval "$_old_shopt_opts"
+  trap - EXIT RETURN
+
+  if [[ $_sourced -eq 1 ]]; then
+    return "$rc"
+  else
+    exit "$rc"
+  fi
+}
+
+if [[ $_sourced -eq 1 ]]; then
+  trap cleanup_shell_state RETURN
+else
+  trap cleanup_shell_state EXIT
+fi
+
+set -e
+
+usage() {
+  cat <<EOF
+Usage: $(basename "$0") OBJECT
+
+Finds the first pkcs11 signing node that works.
+
+This script can either be sourced to populate the environment,
+or executed normally to return json stdout for further processing.
+
+Options:
+  OBJECT: The key that should be found on the node to consider it healthy
+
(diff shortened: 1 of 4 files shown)


scripts/select-pkcs11-node.sh
```diff
@@ -0,0 +1,126 @@
...
+        return 0
+      else
+        jq -n \
+          --arg socket "$socket" \
+          --arg uri "$uri" \
+          '{
+            socket: $socket,
+            token: $token,
+            uri: $uri
+          }'
```
henrirosten, line 116: I believe this fails because jq uses `$token` without `--arg token "$token"`?

Consider:

```
   jq -n \
     --arg socket "$socket" \
     --arg token "$token" \
     --arg uri "$uri" \
     '{
       socket: $socket,
       token: $token,
       uri: $uri
     }'
```

scripts/select-pkcs11-node.sh
```diff
@@ -0,0 +1,126 @@
...
+    uri="pkcs11:token=$token;object=$OBJECT"
+
+    echo "[>] Checking $token on $socket" 1>&2
+    if PKCS11_PROXY_SOCKET="$socket" \
+      timeout "$SOCKET_TIMEOUT" \
+      p11tool \
+      --provider "$PKCS11_PROXY_MODULE" \
+      --login \
+      --list-all \
+      --set-pin "$YUBIHSM_PIN" \
```
henrirosten: `YUBIHSM_PIN` is passed through `p11tool --set-pin`, which exposes the PIN in process arguments while the health check runs. If it's supposed to be secret, consider using `GNUTLS_PIN` or another non-argv mechanism for passing the PIN so it's not exposed e.g. in `ps` listing or similar. 

Caveat: the process may be short-lived, but the script sets a 5s timeout, so failed/unreachable nodes can leave enough time to observe it.

henrirosten (commented):
Not blocking on either if you already considered these

scripts/select-pkcs11-node.sh
joinemm: ah good catch i added it last minute and forgot to push my fix

scripts/select-pkcs11-node.sh
joinemm: resolved

henrirosten (approved):

alextserepov (approved):

fayadf:
@joinemm nethsm gateway server in UAE is behind multiple public ip address and this requires lighthouse to be refreshed with the correct route to the node. This is usually done by the node during restart of tunnel with new handshake but noticed traffic from both gateway IPs at lighthouse, which is unusual causing issues. 

This [PR](https://github.com/tiiuae/ghaf-infra/pull/1051) will force using the static gateway for traffic from the node. Redeployed and tunnel status is up.

PR #1049, opened 2026-05-08
Avoid Jenkins controller executor deadlock

henrirosten:
Fix Jenkins controller executor deadlocks when build pipelines trigger `ghaf-hw-test`, and make UEFI-signing scratch cleanup robust so failed or  aborted runs do not leave large temporary images behind.

Previously, build branches held a `built-in` executor while waiting for the downstream hardware-test job. If enough branches reached that wait at once, all controller executors could be occupied by waiting parent jobs, while `ghaf-hw-test` builds were queued waiting for the same controller executors.

This PR changes the build pipelines to:
  - use `agent none` at the top level
  - allocate `built-in` only for stages that actually need controller workspace access
  - run the blocking `build(job: "ghaf-hw-test", wait: true)` outside `node('built-in')`
  - collect downstream logs/artifacts afterward inside a short controller workspace block
  - use and clean a per-build controller workdir, since `agent none` no longer gives the whole build one reserved Jenkins workspace
  -  move UEFI-signing scratch into a per-build controller workdir and clean it in a `finally` block so failed or aborted runs do not leave large temporary images behind

Tested in ci-dbg and ci-release.

### DIFF
--- a/hosts/hetzci/pipeline-library/vars/utils.groovy
+++ b/hosts/hetzci/pipeline-library/vars/utils.groovy
@@ -253,6 +253,14 @@ def controller_workdir() {
   return "/var/lib/jenkins/ghaf-pipeline-workspaces/${tag}"
 }
 
+def build_tmpdir(String name) {
+  def component = safe_path_component(name)
+  if (!component || component == 'null') {
+    error("Cannot derive a unique build tmpdir for '${name}'")
+  }
+  return "${controller_workdir()}/tmp/${component}"
+}
+
 def clean_controller_workdir() {
   node('built-in') {
     sh "rm -rf '${controller_workdir()}'"
@@ -383,6 +391,7 @@ def create_pipeline(List<Map> targets, String testagent_host = null, String targ
         signed: false,
         reason: null,
         signing_key: null,
+        signing_proxy: null,
       ],
       attestations: [
         provenance: [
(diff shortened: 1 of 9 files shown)


ktusawrk (commented):
**Codex**:
Reviewed for the stated goal: avoiding Jenkins controller executor deadlock.

  The approach looks sound. The key fix is that `build(job: "ghaf-hw-test", wait: true, ...)` now happens outside `node('built-
  in')`, so the parent pipeline no longer holds a controller executor while waiting for the downstream `ghaf-hw-test` job to
  acquire one. The pipeline then reacquires the controller only for the short result/log/artifact collection section.

  I did not find a goal-blocking issue in the PR as written against its original base.

  Residual notes:
  - This avoids the circular wait, but build/provenance/sign/publish work still runs on controller executors, so executor pressure
  can still cause queueing.
  - The PR currently conflicts with `main` in `hosts/hetzci/pipeline-library/vars/utils.groovy`; the resolution should preserve
  both this PR’s executor-release pattern and `main`’s signing redundancy router changes.
  - I ran a Groovy syntax check over the PR checkout with minimal Jenkins stubs; it passed.

henrirosten:
Rebased

ktusawrk (approved):
**Codex**
Looks in line with the goal. The PR removes the top-level built-in agent, scopes controller executor usage to stages that need it, and waits for downstream ghaf-hw-test jobs outside node('built-in'), avoiding the parent/child executor deadlock pattern.

PR #1062, opened 2026-05-15
Balance Nix builder jobs and cores

henrirosten:
Calculate remote-builder `max-jobs` and `cores` from a shared sizing policy. The policy preserves the intended invariant for the configured host sizes: `max-jobs * cores <= vCPU`, but gives more `cores` to each build so the narrow build-graph phases can progress faster. 

The PR also raises Jenkins built-in executors from 4 to 12. This prevents Jenkins from throttling parallel branches before they reach the existing 12-slot `nix-build` lock and Nix builder limits

#### Why?

A build graph is the dependency graph of derivations. Some phases are wide: many independent packages can build in parallel, so high `max-jobs` helps. Other phases are narrow: only one or a few large derivations are ready to be built, often because later work is blocked on them.

In those narrow phases, running many concurrent jobs (`max-jobs`) is not useful because there are not enough independent jobs to fill the builder. Instead, giving each active build more `cores` can help, if those derivations support parallel compilation.

#### What the Values Mean

- `nix.settings.max-jobs`
  Builder-side daemon limit. This is the maximum number of build jobs the remote builder's Nix daemon will actually run concurrently.

- `buildMachines[*].maxJobs`
   Controller-side scheduling limit. This controls how many build jobs a controller may dispatch to a given remote builder.

- `nix.settings.cores`
  Per-build core count. This controls the `NIX_BUILD_CORES` value exposed to each build, allowing individual derivation build jobs to use that many `cores` when the build jobs support parallel builds.

- `hetzci.jenkins.numExecutors`
  Jenkins built-in executor count. This controls how many pipeline branches can run controller-local work concurrently, including the `nix build` steps before Nix/remote-builder throttling takes over.

#### Builder Daemon Limits

| Host | Old `max-jobs` | New `max-jobs` | Old `cores` | New `cores` |
|---|---:|---:|---:|---:|
| `hetzarm` | 40 | 16 | 2 | 5 |
| `hetz86-1` | 48 | 16 | 2 | 6 |
| `hetz86-rel-2` | 48 | 16 | 2 | 6 |
| `hetzarm-rel-1` | 7 | 4 | 2 | 4 |
| `hetzarm-dbg-1` | 7 | 4 | 2 | 4 |
| `hetz86-dbg-1` | 7 | 4 | 2 | 4 |
| `uae-azureci-az86-1` | auto | 6 | 0 | 5 |
| `uae-azureci-hetzarm-1` | 8 | 4 | 2 | 4 |

#### Controller `maxJobs` Limits

| Controller/config | Remote builders | Old `maxJobs` | New `maxJobs` |
|---|---|---:|---:|
| `hetzci-dbg` | arm / x86 | 7 / 7 | 4 / 4 |
| `hetzci-release` | x86 / arm | 48 / 7 | 16 / 4 |
| `hetzci-dev` | arm / x86 | 40 / 48 | 16 / 16 |
| `hetzci-prod` | arm / x86 | 40 / 48 | 16 / 16 |
| `hetzci-vm` | arm / x86 | 40 / 48 | 16 / 16 |
| `hetzci-vm-no-host-nix-store` | arm / x86 | 40 / 48 | 16 / 16 |
| `uae-azureci-prod` | az86 / arm | 16 / 8 | 6 / 4 |
| `hetz86-builder` | hetzarm | 40 | 16 |

#### Jenkins Controller Executors

| Config | Old built-in executors | New built-in executors |
|---|---:|---:|
| all `hetzci.jenkins` environments | 4 | 12 |

Tested in ci-dbg.
Tested in ci-release: https://ci-release.vedenemo.dev/job/ghaf-release-candidate/2/

### DIFF
--- a/hosts/builders/hetz86-1/configuration.nix
+++ b/hosts/builders/hetz86-1/configuration.nix
@@ -11,7 +11,7 @@ let
 
   # Current host sizing: 96 vCPU, 251 GiB RAM, ~1760 GiB /nix disk.
   disk = tuning.mkDiskThresholds 1760;
-  jobs = tuning.mkMaxJobs {
+  build = tuning.mkBuildLimits {
     cpus = 96;
     ramGiB = 251;
   };
@@ -56,8 +56,8 @@ in
     logs.enable = true;
   };
 
-  nix.settings.max-jobs = lib.mkForce jobs;
-  nix.settings.cores = lib.mkForce 2;
+  nix.settings.max-jobs = lib.mkForce build.maxJobs;
+  nix.settings.cores = lib.mkForce build.cores;
   nix.settings.min-free = lib.mkForce disk.minFreeBytes;
   nix.settings.max-free = lib.mkForce disk.maxFreeBytes;
 }
--- a/hosts/builders/hetz86-builder/configuration.nix
+++ b/hosts/builders/hetz86-builder/configuration.nix
@@ -5,8 +5,18 @@
   inputs,
   config,
   machines,
+  lib,
   ...
 }:
+let
+  tuning = import ../../lib/nix-tuning.nix { inherit lib; };
+
+  # Current hetzarm sizing: 80 vCPU, 250 GiB RAM.
+  armBuilder = tuning.mkBuildLimits {
+    cpus = 80;
+    ramGiB = 250;
+  };
+in
 {
   imports = [
     ./disk-config.nix
@@ -49,7 +59,7 @@
       {
         hostName = "hetzarm.vedenemo.dev";
         system = "aarch64-linux";
-        maxJobs = 40;
+        maxJobs = armBuilder.maxJobs;
         speedFactor = 1;
         supportedFeatures = [
           "nixos-test"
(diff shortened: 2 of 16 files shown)


ktusawrk (approved):
Let's keep an eye how this impacts to the server performance.

Here is Codex response to a question if 'light' oversubscription would helpful:

>Oversubscription
Light oversubscription could help, but I would not bake it into this PR without measurements. The current invariant max-jobs * cores <= vCPU is a good first step because it reduces memory pressure while giving narrow graph phases more per-derivation parallelism.
>
>Where oversubscription might help: derivations that do not actually consume all NIX_BUILD_CORES, I/O-heavy phases, or hosts showing sustained CPU idle during builds.
>
>Where it can hurt: link-heavy C/C++/Rust builds, JVM builds, tests, and nixos-test/KVM workloads. It also lowers memory per active job, which is especially risky on the 16-vCPU/30-GiB builders.
>
>If experimenting, I’d start only on the large builders, with something like 10-15% aggregate oversubscription, and compare CPU utilization, memory PSI/swap/OOMs, queue time, and wall-clock pipeline duration. I would avoid oversubscribing the small 16-vCPU builders initially.

PR #1074, opened 2026-05-21
Add fleetdm server nix configuration

joinemm:
Adds fleetdm server configured in nix. It should match the default docker-compose stack from upstream fleetdm, just re-implemented using systemd services and no docker.
- Mysql
- Redis
- Fleetdm
- Nginx

The api key and enrollment token has changed as result. Those secrets have been updated.

Fleet package was vendored in from nixpkgs-unstable, as the version from our pinned nixpkgs is outdated, and does not include the frontend assets necessary to run the web interface. This can be removed once we update to nixos 26.05.

### DIFF
--- a/hosts/ghaf-fleetdm/fleet.nix
+++ b/hosts/ghaf-fleetdm/fleet.nix
@@ -1,0 +1,197 @@
+# SPDX-FileCopyrightText: 2026 TII (SSRC) and the Ghaf contributors
+# SPDX-License-Identifier: Apache-2.0
+{
+  config,
+  pkgs,
+  self,
+  ...
+}:
+
+let
+  domain = "fleetdm.vedenemo.dev";
+  fleetPort = 1337;
+  inherit (self.packages.${pkgs.stdenv.hostPlatform.system}) fleet fleetctl;
+in
+{
+  environment.systemPackages = [
+    fleet
+    fleetctl
+  ];
+
+  users.groups.fleet = { };
+
+  users.users.fleet = {
+    isSystemUser = true;
+    group = "fleet";
+    home = "/var/lib/fleet";
+    createHome = true;
+  };
+
+  sops.secrets."fleet-private-key" = {
+    owner = "fleet";
+    group = "fleet";
+    mode = "0400";
+  };
+
+  sops.templates."fleet-env" = {
+    owner = "fleet";
+    group = "fleet";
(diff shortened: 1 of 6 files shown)


hosts/ghaf-fleetdm/fleet.nix
```diff
@@ -0,0 +1,181 @@
...
+      username: fleet
+
+    redis:
+      address: 127.0.0.1:${toString config.services.redis.servers.fleet.port}
+      database: 0
+
+    server:
+      address: 127.0.0.1:${toString fleetPort}
+      tls: false
+      private_key_path: ${config.sops.secrets."fleet-private-key".path}
```
henrirosten: `server.private_key_path` is not a supported Fleet config key. Fleet only reads `server.private_key`, so this value will be ignored and the service
  will start with `server.private_key` unset.

I believe the key needs to be injected as `server.private_key` or `FLEET_SERVER_PRIVATE_KEY`, not `private_key_path`.

joinemm: You're right. I refactored it to use FLEET_SERVER_PRIVATE_KEY

hosts/ghaf-fleetdm/fleet.nix
```diff
@@ -0,0 +1,196 @@
...
+      username: fleet
+
+    redis:
+      address: 127.0.0.1:${toString config.services.redis.servers.fleet.port}
+      database: 0
+
+    server:
+      address: 127.0.0.1:${toString fleetPort}
+      tls: false
+
```
henrirosten, line 87: `server.trusted_proxies` is left unset here, so Fleet falls back to its legacy client-IP behavior: it trusts `True-Client-IP`, then `X-Real-IP`, then `X-Forwarded-For`. Since this host is internet-facing through nginx, and nginx sets `X-Real-IP` but does not clear client-supplied `True-Client-IP`, external clients can spoof the client IP Fleet records and uses for IP-based failed-request banning/rate limiting.

A simple fix would be:
```
      server:
        address: 127.0.0.1:${toString fleetPort}
        tls: false
        trusted_proxies: "header:X-Real-IP"
```

joinemm: Unlikely to be a real issue but fixed anyway

henrirosten (approved):

PR #1078, opened 2026-05-22
Upload test results alongside OCI artifact

joinemm:
Added a subcommand to `oci-publish` which zips up and uploads the test-results directory to the OCI registry, attaching it to the main artifact. This runs as a separate step after the HW testing has completed.

Test run: https://ci-dbg.vedenemo.dev/job/ghaf-manual/27/

### DIFF
--- a/hosts/hetzci/pipeline-library/vars/utils.groovy
+++ b/hosts/hetzci/pipeline-library/vars/utils.groovy
@@ -545,26 +545,19 @@ def create_pipeline(List<Map> targets, String testagent_host = null, String targ
         }
         if (!no_image) {
           stage("Publish OCI ${shortname}") {
-            if (sh(
-              script: 'command -v oci-publish >/dev/null 2>&1',
-              returnStatus: true
-            ) != 0) {
-              println("Skipping OCI publish ${shortname}: oci-publish is not installed")
-            } else {
-              withCredentials([string(credentialsId: 'oci_registry_password', variable: 'OCI_PASSWORD')]) {
-                def job_name = env.JOB_BASE_NAME.replaceFirst('^ghaf-', '')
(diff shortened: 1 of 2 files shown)


hosts/hetzci/pipeline-library/vars/utils.groovy
```diff
@@ -609,6 +602,25 @@ def create_pipeline(List<Map> targets, String testagent_host = null, String targ
...
+                script: "test -d '${output}/test-results'",
+                returnStatus: true
+              ) != 0) {
+                error("Missing test results for ${shortname}: ${output}/test-results")
+              }
+              withCredentials([string(credentialsId: 'oci_registry_password', variable: 'OCI_PASSWORD')]) {
+                sh """
+                  oci-publish test-results \
+                    --results-dir '${output}/test-results' \
+                    --subject-reference '${oci_result.primary.reference}'
```
henrirosten, line 618: Passing `oci_result.primary.reference` to `oci-publish test-results` breaks `OCI_LAYOUT_PATH` mode because the existing referrer path uses the tag reference there (`pkgs/oci-publish/src/oci_publish.py:261`); ORAS cannot resolve the digest reference in a local OCI layout.

One way to fix this: either pass `oci_result.primary.tag_reference` only when `OCI_LAYOUT_PATH` is set, or preferably move this decision into `oci-publish test-results` so callers do not need to know about layout-vs-registry subject handling. For example, the CLI could accept both references or a result JSON path and choose `primary.tag_reference` when `OCI_LAYOUT_PATH` is set, otherwise `primary.reference`.

joinemm: OCI_LAYOUT_PATH mode is only used for debugging so jenkins implementation not supporting it doesn't matter.

henrirosten (approved):

PR #1080, opened 2026-05-22
Redeploy testagent prod in UAE

fayadf:
Testagent prod machine had to be redployed in uae lab due to faulty usb port.
Cleaned up sops as nebula tunnel was removed from node1 in lab but was still allowed to decrypt ca. 
Hetzarm was redeployed so ssh keys updated.
Jenkins plugin is now reused from hetzci.
Netvm hostname and device-id has been updated on agx32 target.
Reconfigure registry to use filesystem for storage as S3 in uae me-central-1 is unreliable after the recent incident.


fayadf:
Registry will use filesystem for storage because S3 in uae me-central-1 is unreliable after the incident.

joinemm (approved):

PR #1099, opened 2026-06-03
Pull artifacts from OCI registry on release publish

joinemm:
Use the new ghaf-fetch script to download the artifacts from the OCI registry and then upload them into ghaf archive.

Artifacts dir is still required for OTA pinning to stay functional.

The script checks if any release targets exist with the given tag and downloads the ones that do, this way it works for laptop-only releases as well.

### DIFF
--- a/hosts/hetzci/pipelines/ghaf-release-publish.groovy
+++ b/hosts/hetzci/pipelines/ghaf-release-publish.groovy
@@ -8,10 +8,17 @@ properties([
       Ghaf release tag or version name used to identify the release in the archive. Example: 'ghaf-25.12.1'
       '''.stripIndent()),
     string(name: 'ARTIFACTS_URL', defaultValue: '', description: '''
-      Specify the artifacts URL from where the release artifacts will be read.
+      Specify the artifacts URL used for local artifact lookup and OTA pinning.
+      If OCI_TAG is empty, release artifacts are also archived from this local artifact path.
       If left empty, uses the current latest artifacts from the 'ghaf-release-candidate' pipeline.
       Example:
       'https://ci-release.vedenemo.dev/artifacts/ghaf-release-candidate/20251210_081817797-commit_28fb2bdcbb558d02c33b01ef25a2250ff3fdc479/'
+      '''.stripIndent()),
+    string(name: 'OCI_TAG', defaultValue: '', description: '''
+      Specify an OCI registry tag.
+      If set, all release targets are archived from the OCI registry using this tag. ARTIFACTS_URL must be set and is still used for OTA pinning.
+      Example:
+      'release-20260603_123242705-29e9ef6e98ef9a589b5814fe438ba94601df1bdc'
       '''.stripIndent()),
     string(name: 'BUCKET', defaultValue: null, description: '''
       Override the object storage bucket to push to, leave empty to select automatically. Example: 'ghaf-artifacts-dev'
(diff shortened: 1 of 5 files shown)


scripts/archive-ghaf-release.sh
```diff
@@ -180,22 +214,69 @@ verify_signatures() {
...
+    oci_target_name="${target_name#packages.}"
+    target_reference="$OCI_REPOSITORY_PREFIX/$oci_target_name:$tag"
+    target_dir="$artifactsdir/$target_name"
+
+    mkdir -p "$target_dir"
+    echo "[+] Pulling release artifacts from OCI: $target_reference"
+    if ! (
+      cd "$target_dir"
+      ghaf-fetch --all "$target_reference"
+    ); then
```
henrirosten, line 252: This skips on any `ghaf-fetch` failure, not just "target/tag not found". A transient registry/ORAS error would be treated the same as an intentionally absent target and could still produce a partial release. Can we fail here for unexpected pull errors and only skip genuine not-found cases?

scripts/archive-ghaf-release.sh
```diff
@@ -180,22 +214,69 @@ verify_signatures() {
...
+  if [[ $tag == */* || $tag == *@* || $tag == *:* ]]; then
+    print_err "OCI release archival requires a tag, not a reference: $tag"
+    exit 1
+  fi
+
+  mkdir -p "$artifactsdir"
+  pulled_targets=0
+  for target_name in "${release_targets[@]}"; do
+    oci_target_name="${target_name#packages.}"
+    target_reference="$OCI_REPOSITORY_PREFIX/$oci_target_name:$tag"
```
henrirosten, line 231: `oci-publish` lowercases repository names before pushing, but this fetch path preserves the mixed-case target name. For `...intel-laptop-storeDisk-debug-installer` that points at `.../storeDisk...`, while the live registry repo is `.../storedisk...`. In OCI archive mode this target will be skipped even when it exists. Can we normalize `oci_target_name` to lowercase before building `target_reference`?

joinemm:
Follow-up: I moved final test metadata to `test-results.json` only, and included that file in the OCI test-results artifact.

I also removed the post-test `manifest.json` patching. Since `manifest.json` is already used as the OCI config when the target artifact is pushed, editing it afterwards would make the local file differ from what was actually published to OCI.

hosts/hetzci/pipelines/ghaf-release-publish.groovy
```diff
@@ -108,9 +119,15 @@ pipeline {
                 } else {
                   env.BUCKET=params.BUCKET
                 }
-                sh """
-                  archive-ghaf-release -a "$ARTIFACTS_DIR" -t "${params.GHAF_VERSION}"
-                """
+                if (env.OCI_TAG) {
+                  sh """
+                    archive-ghaf-release -o "$OCI_TAG" -t "${params.GHAF_VERSION}"
+                  """
```
henrirosten, line 128: I think it would be safer to require `ARTIFACTS_URL` when `OCI_TAG` is set. Right now OTA pinning still uses the local `ARTIFACTS_DIR`, while archiving switches to OCI here, so the job can end up mixing artifacts from two different builds.

A simple fix would be to error out early if `OCI_TAG` is set but `ARTIFACTS_URL` is not. That would make the expected input explicit and avoid silently publishing a mixed release.

henrirosten (approved):

PR #1102, opened 2026-06-04
ci: make Acroname SSD reconnect more robust

ktusawrk:
## Summary

  Make hardware flashing more robust when reconnecting Jetson AGX SSDs through the Acroname programmable USB hub.

  The previous flow switched the hub mode, slept for a fixed 10 seconds, and checked `/dev/ssdORINAGX1` once. If USB enumeration or udev symlink creation was late, the job
  failed with:

  `Symlink /dev/ssdORINAGX1 not found. Failed to connect target USB disk to test agent.`

  ## Changes

  - Run `udevadm settle` after Acroname/usbsdmux mode switches.
  - Retry target USB disk connection up to 3 times.
  - Wait for the SSD symlink to appear instead of checking only once.
  - Toggle the hub back to DUT mode between failed connection attempts.
  - Wait for the SSD symlink to disappear during teardown.
  - Print `lsblk` and recent USB/storage `dmesg` output when reconnect still fails.

  ## Risk / Rollback

  Risk is limited to Jenkins hardware-test flashing setup/teardown.

  This may make real SSD/hub failures take longer to fail, but should reduce false failures caused by slow USB enumeration or udev races. Rollback is to revert this commit.

  ## Validation

  - `git diff --check`
  - `nix fmt`
  - `nix flake check --no-build --option allow-import-from-derivation false`

### DIFF
--- a/hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy
+++ b/hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy
@@ -110,42 +110,79 @@ def setup_mount_commands(String conf_file_path, String target, String device_nam
   if (target.contains("microchip-icicle-")) {
     def muxport = get_test_conf_property(conf_file_path, device_name, 'usb_sd_mux_port')
     return [
-      mount_cmd: "/run/wrappers/bin/sudo usbsdmux ${muxport} host; sleep 10",
-      unmount_cmd: "/run/wrappers/bin/sudo usbsdmux ${muxport} dut",
+      mount_cmd: "(/run/wrappers/bin/sudo usbsdmux ${muxport} host; rc=\$?; /run/wrappers/bin/sudo udevadm settle --timeout=10 || true; sleep 2; exit \$rc)",
+      unmount_cmd: "(/run/wrappers/bin/sudo usbsdmux ${muxport} dut; rc=\$?; /run/wrappers/bin/sudo udevadm settle --timeout=10 || true; sleep 2; exit \$rc)",
     ]
   }
   def serial = get_test_conf_property(conf_file_path, device_name, 'usbhub_serial')
   return [
-    mount_cmd: "/run/wrappers/bin/sudo AcronameHubCLI -u 0 -s ${serial}; sleep 10",
-    unmount_cmd: "/run/wrappers/bin/sudo AcronameHubCLI -u 1 -s ${serial}; sleep 10",
+    mount_cmd: "(/run/wrappers/bin/sudo AcronameHubCLI -u 0 -s ${serial}; rc=\$?; /run/wrappers/bin/sudo udevadm settle --timeout=10 || true; sleep 2; exit \$rc)",
+    unmount_cmd: "(/run/wrappers/bin/sudo AcronameHubCLI -u 1 -s ${serial}; rc=\$?; /run/wrappers/bin/sudo udevadm settle --timeout=10 || true; sleep 2; exit \$rc)",
   ]
 }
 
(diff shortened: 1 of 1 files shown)


hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy
```diff
@@ -110,42 +110,71 @@ def setup_mount_commands(String conf_file_path, String target, String device_nam
...
-      echo "Symlink ${dev} not found. Failed to connect target USB disk to test agent."
-      echo "Check USB cables. Maybe need to reboot test agent or Acroname USB hub."
-      echo "Aborting flashing ${device_name}"
-      exit 1
+      /run/wrappers/bin/sudo rm ${dev}
     fi
+
+    for attempt in \$(seq 1 3); do
+      echo "Connecting target USB disk to test agent, attempt \${attempt}/3..."
+      ${mount_cmd}
```
cazfi: There was 'set +e' added before this, so mount command failure here is now fatal. Is that what we want?

hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy
```diff
@@ -110,42 +110,71 @@ def setup_mount_commands(String conf_file_path, String target, String device_nam
...
+        /run/wrappers/bin/sudo udevadm settle --timeout=2 || true
+        if /run/wrappers/bin/sudo test -L ${dev}; then
+          echo "Symlink ${dev} found."
+          exit 0
+        fi
+        sleep 2
+      done
+
+      echo "Symlink ${dev} not found after attempt \${attempt}; toggling USB hub back to target before retry."
+      ${unmount_cmd}
```
cazfi: If there's an issue with the mount, unmounting certainly fails (cannot unmount what is not mounted). Together with 'set +e' this exits rather than continues to the next attempt in the loop.

cazfi (changes requested):

hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy
ktusawrk: Good point. The intention was to make the retry loop tolerant of transient hub/udev state, not to make all command failures silently non-fatal.

  With `set +e`, the current script keeps going after `${mount_cmd}` or `${unmount_cmd}` failures, which means the retry loop can continue, but it also hides real Acroname/
  usbsdmux command failures. That is too broad.

  I’ll update this to avoid global `set +e` and handle the expected retry cases explicitly:
  - keep `test -L` checks inside `if` statements
  - treat a failed `${mount_cmd}` as a failed attempt, log it, toggle back if possible, then continue/retry
  - make the retry cleanup `${unmount_cmd}` non-fatal with logging, because if host-side connect failed the DUT-side switch may also fail or be unnecessary
  - keep the final failure path with `lsblk`/`dmesg` diagnostics

  That should make the retry behavior explicit instead of relying on shell error-mode side effects.

hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy
ktusawrk: Yes, this cleanup path should not abort the retry loop. If the host-side connect did not complete, switching back to target side is best-effort cleanup before the next
  attempt. I’ll make that explicit with `if ! ${unmount_cmd}; then ...; fi` rather than relying on `set +e`.

ktusawrk:
@cazfi Improvements based on your review comments are in (set +e is now gone), please re-review.

cazfi (approved):

ktusawrk:
Tested by running ghaf-main in dev. All SSD drives were connected successfully on the first try.

PR #1105, opened 2026-06-07
grafana: Use explicit secret_key

cazfi:


henrirosten (approved):
Change looks good, approved.

For future PRs, could you please add a short description of the change and the testing you did?

PR #1114, opened 2026-06-11
Update registry access control to allow authenticated pushes

joinemm:
Allow authenticated developers to create a zot api key and push artifacts to the registry into arbitrary repositories, but not under `ghaf/`; that remains jenkins-only.

### DIFF
--- a/hosts/ghaf-registry/configuration.nix
+++ b/hosts/ghaf-registry/configuration.nix
@@ -77,6 +77,8 @@ let
       auth = {
         htpasswd.path = config.sops.secrets.zot-htpasswd.path;
 
+        apikey = true;
+
         openid.providers.oidc = {
           name = "Vedenemo Auth";
           issuer = "https://auth.vedenemo.dev";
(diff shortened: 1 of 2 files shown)


hosts/ghaf-registry/configuration.nix
```diff
@@ -110,7 +110,12 @@ let
             anonymousPolicy = [ "read" ];
           };
           "**" = {
-            defaultPolicy = [ "read" ];
+            defaultPolicy = [
```
henrirosten, line 115: This widens write access for authenticated users, but it does not enable the zot API-key flow described in the PR body. The registry `auth` config still only sets `htpasswd` and `openid`; without `apikey = true`, `/zot/auth/apikey` is not available. So this appears to implement the permission change, but not the “developers can create a zot API key” part.

joinemm: not sure what this auth.apikey option actually does but right now it's possible to create a key and use it as basic auth password for pushing. Still, for completeness i've added it now.

hosts/ghaf-registry/configuration.nix
```diff
@@ -110,7 +112,12 @@ let
...
+              "read"
+              "create"
+              "update"
+              "delete"
```
henrirosten, line 119: This seems broader than “allow authenticated pushes” you mention in the description: adding `delete` here also allows authenticated users to delete artifacts in non-`ghaf/**` repositories. Is that intentional?

joinemm: If they push something they should be able to delete it too

henrirosten (approved):

PR #1120, opened 2026-06-16
Document proper way to do deployment over release boundary

cazfi:
Deployment with 'switch' might fail when deploying over nixos release boundary. Deploy with --boot instead.

### DIFF
--- a/docs/deploy-rs.md
+++ b/docs/deploy-rs.md
@@ -38,3 +38,19 @@ deploy -si --ssh-user myuser .#ghaf-monitoring
 ```
 
 Running just `deploy` without any arguments will deploy every host defined in the configuration, even if they have no changes. This is usually not something you want to do.
+
+## Release upgrades
+
+Do not use a normal deploy with immediate activation ('switch') when the target host is still running a different NixOS release than the configuration being deployed, for example `25.11 -> 26.05`.
+
+Those upgrades can fail during activation while reloading systemd user units.
+
+Use a boot deployment instead:
+
+```sh
+deploy --boot .#testagent-release
+```
+
+Then reboot the host into the newly deployed generation.
+
+Note that there is no automatic rollback on the boot deployment.


ktusawrk (commented):
The new `Release upgrades` section makes sense overall. Using `deploy --boot` across a NixOS release boundary is a reasonable workaround when immediate activation can fail.

  I think one important detail is still missing: this changes the failure mode compared to a normal `deploy`. Earlier in this doc we mention deploy-rs automatic rollback, but with `--boot` the new generation is only activated on reboot, so a bad upgrade may fail at boot time instead of during deployment, and recovery may be manual. That tradeoff should be stated explicitly so operators do not assume the same rollback behavior still applies.

  I would also tighten the wording slightly:

  - replace `deploy/switch` with something like `plain deploy with immediate activation`
  - use a generic example such as `deploy --boot .#<host>` unless `testagent-release` is intentionally the canonical case

cazfi:
> 
> I think one important detail is still missing: this changes the failure mode compared to a normal `deploy`. Earlier in this doc we mention deploy-rs automatic rollback, but with `--boot` the new generation is only activated on reboot, so a bad upgrade may fail at boot time instead of during deployment, and recovery may be manual. That tradeoff should be stated explicitly so operators do not assume the same rollback behavior still applies.

Added a note about automatic rollback not being active.

> I would also tighten the wording slightly:
> 
>     * replace `deploy/switch` with something like `plain deploy with immediate activation`

Reworded that part slightly.

>     * use a generic example such as `deploy --boot .#<host>` unless `testagent-release` is intentionally the canonical case

Using testagent-release host name verbatim in the example is in line with the other examples in the same document. While it would make sense to have them all replaced with a <host> placeholder, that should be a separate document-wide change.

joinemm (approved):

ktusawrk (approved):

PR #1126, opened 2026-06-22
Use environment specific mutable tags for all pipelines

joinemm:
Pipelines now tag artifacts as `${env}-latest` during OCI publish step. This tag will always point to the latest build for the target in this pipeline.

Additionally, tag release builds when they are published with the ghaf version.

### DIFF
--- a/hosts/uae/azureci/registry/configuration.nix
+++ b/hosts/uae/azureci/registry/configuration.nix
@@ -42,6 +42,12 @@ let
             deleteUntagged = true;
             keepTags = [
               {
+                patterns = [
+                  ".*-latest"
+                  "ghaf-.*"
+                ];
+              }
+              {
                 patterns = [ "release-.*" ];
                 # Keep at least three builds in each repository
                 mostRecentlyPushedCount = 3;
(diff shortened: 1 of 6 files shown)


hosts/uae/azureci/registry/configuration.nix
```diff
@@ -41,6 +41,12 @@ let
             deleteReferrers = true;
             deleteUntagged = true;
             keepTags = [
+              {
+                patterns = [
+                  ".*-latest"
+                  "ghaf-.*"
```
cazfi, line 47: Does this keep release builds infinitely in the registry? Is that an intentional change to the behavior?

cazfi (changes requested):

hosts/uae/azureci/registry/configuration.nix
joinemm: I think we should keep published release builds infinitely in the registry, especially if we want to make the registry the main source to get ghaf releases at some point.

cazfi: If we make the registry the main source, yes. Are our registry machines prepared to hold full set of builds generated every two weeks (26 sets a year)? I though that limited space was the reason we clean other builds away so aggressively.

joinemm: we are using autoscaling s3 object storage so the only limitation is how much we want to spend

cazfi (approved):

PR #1131, opened 2026-06-25
Add nethsm-gateway-dev config

joinemm:
NixOS configurations for a second nethsm in Tampere lab, to be used as dev/testing setup. The existing gateway was moved under prod/

Also refactored the uae gateway to reuse the same modules.


fayadf (approved):

hosts/nethsm-gateway/dev/configuration.nix
```diff
@@ -0,0 +1,25 @@
...
+{ ... }:
+{
+  imports = [
+    ./disk-config.nix
+    ../common.nix
+  ];
+
+  sops.defaultSopsFile = ./secrets.yaml;
+
+  system.stateVersion = "25.11";
```
cazfi, line 12: If this lands only after ghaf-infra 26.05 update ( #1118 ), should be "26.05".

joinemm: I've installed the host in question already at this point and I will not reinstall it when/if 26.05 lands (only deploy)

PR #1132, opened 2026-06-29
Add release policy with checker and attestation file

joinemm:
Run policy checker in release mode at the end of release candidate pipeline. It checks the items defined in release policy file and outputs a signed attestation file with fail/pass fields and relevant subject data.

### DIFF
--- a/slsa/release-policy.yaml
+++ b/slsa/release-policy.yaml
@@ -1,0 +1,41 @@
+version: '1.0'
+name: Ghaf Release Policy
+uri: https://github.com/tiiuae/ghaf-infra/blob/main/slsa/release-policy.yaml
+
+criteria:
+  - id: release-builder
+    description: Artifact provenance identifies a release CI builder
+    required: true
+    cel: |
+      files.provenance &&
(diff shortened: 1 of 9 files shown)


slsa/release-policy.yaml
```diff
@@ -0,0 +1,41 @@
...
+    cel: provenance_policy.passed
+
+  - id: sbom-present
+    description: SBOM outputs are present
+    required: true
+    cel: files.sbom_csv && files.sbom_cyclonedx && files.sbom_spdx
+
+  - id: hardware-tests
+    description: Configured hardware tests completed successfully or were explicitly skipped
+    required: true
```
cazfi: This is quite strict, In the past we've deemed some failures acceptable afterwards, especially when the failure has been due to hardware malfunction or similar.

joinemm: fair, perhaps we could make it run but not required

cazfi (dismissed):

joinemm:
had to update this now that https://github.com/tiiuae/ghaf-infra/pull/1135 landed

joinemm:
fixed conflicts

PR #1158, opened 2026-07-20
Refactor builders, testagents and registry configs

joinemm:
Probably easier to review the commits separately.

- Move remote builders "tuning" into a module with options instead of repeating the lib function calls and assignments on each builder.
- Use shared files for uae and finland testagents, with credential file as an option. Remove duplicate `agents-common.nix`
- Use shared registry module for ghaf-registry and uae registry.
- Fixed regex escapes. nix will simply remove the single backslashes resulting in a non-escaped period.


hosts/registry.nix
```diff
@@ -0,0 +1,228 @@
...
+          proxy_read_timeout 3600s;
+          proxy_send_timeout 3600s;
+          client_body_timeout 3600s;
+        '';
+      };
+
+      locations."/metrics" = lib.mkIf cfg.metrics.enable {
+        proxyPass = "http://127.0.0.1:${toString zotPort}/metrics";
+        extraConfig = ''
+          allow ${machines.ghaf-monitoring.internal_ip};
```
henrirosten, line 191: The shared module now enables the UAE registry `/metrics` endpoint with an allowlist for `ghaf-monitoring.internal_ip` (`10.0.0.2`). That IP is the Hetzner monitoring address, not UAE-specific. If UAE registry metrics are meant to be scraped, this allowlist should probably be configurable.

hosts/registry.nix
```diff
@@ -0,0 +1,228 @@
...
+          clientsecret = config.sops.placeholder.auth-client-secret;
+        };
+      };
+    };
+
+    services.nginx.virtualHosts.${cfg.domain} = {
+      enableACME = true;
+      forceSSL = true;
+      default = true;
+      http2 = false;
```
henrirosten: This changes UAE registry from the nginx default `http2 = true` to `false`. If intentional, fine; otherwise `http2` should probably be configurable or set only for `ghaf-registry`.

henrirosten (commented):
These comments do not block the merge, but still I think worth a comment.

henrirosten (dismissed):

hosts/registry.nix
joinemm: fair point but i'd rather address it when/if uae needs metrics and we know what kind of allowlist it needs.

hosts/registry.nix
joinemm: Changed this to derive its value from whether s3 storage is configured or not

henrirosten (approved):

PR #1170, opened 2026-07-22
Standardize all hosts to use systemd-boot

joinemm:
ALl hosts now use systemd-boot instead of grub and the config lives in one place in `hosts/common.nix`

when deploying a bootloader change the following command should be used instead of deploy-rs:

```
nixos-rebuild switch --flake .#HOST --target-host HOST --use-remote-sudo --install-bootloader
```

### DIFF
--- a/hosts/common.nix
+++ b/hosts/common.nix
@@ -21,6 +21,13 @@ in
     "ghaf-dev"
   ];
 
+  boot.loader.systemd-boot = {
+    enable = true;
+    configurationLimit = 3;
+  };
+
+  boot.loader.timeout = 1;
+
   # revision of the flake the configuration was built from.
   # $ nixos-version --configuration-revision
   system.configurationRevision = toString (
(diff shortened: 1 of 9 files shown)


hosts/common.nix
```diff
@@ -21,6 +21,13 @@ in
     "ghaf-dev"
   ];
 
+  boot.loader.systemd-boot = {
+    enable = true;
+    configurationLimit = 3;
+  };
```
henrirosten, line 27: This does not preserve the previous `boot.loader.efi.canTouchEfiVariables = true` from `testagent/agents-common.nix` and `nethsm-gateway/common.nix`. Is that intentional?

Consider either setting this in the new common bootloader block:

```nix
  boot.loader.efi.canTouchEfiVariables = true;
```
or restore it in the affected host-family modules if it should not be global?

joinemm: I'm not sure if that's actually required for anything, but I restored it for those lab hosts that had it set before

henrirosten (approved):

PR #1181, opened 2026-07-27
Support DarterSB-prod

azbeleva:
Add Darter secure-boot support for `intel-laptop-debug` tests by enabling it in the manual pipeline, wiring `darter-pro` to `darter-sec-boot` in the pipeline model, and updating Darter-Secure-Boot configuration.

### DIFF
--- a/hosts/hetzci/pipeline-library/vars/pipelineModel.groovy
+++ b/hosts/hetzci/pipeline-library/vars/pipelineModel.groovy
@@ -95,7 +95,12 @@ private def display_device_tag(Map testRun) {
 
   def explicitTag = normalize_optional_string(testRun.get('device_tag', null))
   if (explicitTag != null) {
-    return explicitTag == 'x1-sec-boot' ? 'lenovo-x1' : explicitTag
+    if (explicitTag == 'x1-sec-boot') {
+      return 'lenovo-x1'
+    } else if (explicitTag == 'darter-sec-boot') {
+      return 'darter-pro'
+    }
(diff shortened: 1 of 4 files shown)


hosts/hetzci/pipeline-library/vars/pipelineModel.groovy
```diff
@@ -196,6 +206,10 @@ private def inferred_device_info(String targetName, boolean secureboot) {
     return info_for_device_tag('x1-sec-boot')
   }
 
+  if (normalizedTarget.contains('system76-darp11-b') && secureboot) {
+    return info_for_device_tag('darter-sec-boot')
+  }
+
```
henrirosten, line 212: This new `system76-darp11-b` secure-boot mapping will produce `DEVICE_TAG=darter-sec-boot`, but `hwTestUtils.extra_tag_suffix()` only adds `NOTexcl-secboot` when the target contains "darter". Since these targets are named `system76-darp11-b-*`, the secure-boot exclusion filter will not be applied.

I believe you also need to update `hosts/hetzci/pipeline-library/vars/hwTestUtils.groovy:38`:
Change:
```
  if (target.contains("darter")) {
```
to something like:
```
  if (target.contains("darp11-b")) {
```

azbeleva: Good catch. I did not check that, it was done before. Fixed now.

henrirosten (approved):

PR #1189, opened 2026-07-28
Refactor host configuration into reusable NixOS modules

joinemm:
- Move shared, environment, ZRAM, and Zot registry configuration from `hosts/` into reusable NixOS modules.
- Pin `system.stateVersion` explicitly per host instead of providing a common default.
- Generalize Zot registry overrides with `extraConfig`.
- Relocate `services-*` under new `modules/` layout.
- Remove legacy `service-binary-cache`, `service-rclone-http` and `service-remote-build` as those are not used by any host.

Consider this a phase 1 of a more substantial refactoring. If it proves to work well, we should move more code under `nixosModules`


modules/environment/hetzner-robot.nix
```diff
@@ -23,21 +15,17 @@
     auth.password_file = config.sops.secrets.loki_password.path;
   };
 
-  hardware.enableRedistributableFirmware = true;
-
-  networking.useNetworkd = true;
-  networking.useDHCP = false;
-  networking.usePredictableInterfaceNames = false;
+  networking = {
```
henrirosten, line 19: The refactor drops `efibootmgr` from Hetzner Robot hosts. The old `hosts/hetzner-robot.nix` added it to `environment.systemPackages`, but the new module no longer does; `hetz86-1.config.environment.systemPackages` now evaluates without it.

Was this intentional?

joinemm: what is the use case for having it there?

henrirosten: It was added in this PR: https://github.com/tiiuae/ghaf-infra/pull/998 which hints the use-case too.
But I have no idea if there are any current use-cases, just pointing out the diff.

joinemm: Added efibootmgr back

henrirosten (approved):

PR #1195, opened 2026-07-30
Revert to GRUB on hetzner cloud hosts that don't have UEFI support

joinemm:
Hetzner Cloud hosts do not all use the same firmware mode. Some boot with UEFI, while others use legacy BIOS.

Hetzner documents that newer server types use UEFI by default but this does not apply retroactively to every existing Cloud instance or older server type. See https://docs.hetzner.com/robot/dedicated-server/operating-systems/uefi/.

Each new cloud instance should be verified for EFI support (with `bootctl` or something else)

Enabling systemd-boot globally caused legacy BIOS hosts to become unbootable once grub files were deleted from the system, because systemd-boot requires UEFI. `ghaf-webserver` had to be reinstalled after encountering this issue.

### Changes

- Add `virtualisation.hetzner.withEfiSupport` to the Hetzner Cloud module.
  - Defaults to `false`.
  - Legacy hosts use GRUB.
  - UEFI hosts use systemd-boot.
- Enable UEFI support explicitly for the verified UEFI Cloud hosts:
  - `ghaf-registry`
  - `hetzarm-dbg-1`
  - `hetzarm-rel-1`
- Keep systemd-boot as the default for non-Hetzner-Cloud hosts.
- Add `hcloud` to the development shell for inspecting Hetzner Cloud server metadata.

Hetzner Robot hosts are unaffected by the new option. Their boot mode was verified separately, and all currently use UEFI.


hosts/ghaf-webserver/secrets.yaml
```diff
@@ -1,30 +1,44 @@
 ssh_host_ed25519_key: ENC[AES256_GCM,data:LrX77m0elP4dxO0MlG4d3ZDUUsstvSUHOGxmCezoBmqMIo6lwj+8edMUrfKwiEQeYiOwZvInqZaQkjwYu/7q13tHX+P0VDdwNR51P8ot3QzCPxA4f6cXoMzYo/Bg+VtIDw1IdN87yM0Q8tLSMJNckvK1wQQbNDGann1smaSse6PWQ3kQ2M/lv3dud9RcIPZjFYXUvmYQQjrGWEyoQsMhQqUsXktk2n363uofBec0Jk2RwBjpV29wX6zhAXo1wny23XCRDLSltUhcvjPS8E/uutAJDnVgnLBRpPn1BqBIihRfnYTavD9Ru9P1iqGEEcLUgmxN/I7HlzaIzIbmbGbp/JDkGCeVisbCtID2opVAd+TnACZvQYi4awSJLTEjz3alP9X2lFAq5qhqBdk7449f41apqwjfFufZN1ZL0OlS70L74O1av46ZIt/3kvXo7YwMrHaJgtYfRzepxAitSeEh2upW1qG4XzfRqbzgHQUp7QV0YWC3Sst+OPKe5IENycReeedqr1x5AMjs5bOklv6Y9qskj8AVu1J4ys6D,iv:yx2QLsp7WD7Z1DPtfEirhkARukNIokqKCf62XSWow5s=,tag:ejZphUeRP9pkCmHrT93jIQ==,type:str]
+something: ENC[AES256_GCM,data:0lGv,iv:hy1aziEqfCpNbG1Y9r6jmqIIAuNKys0QUnzt8LvkTKc=,tag:Q/vpx1r22cTbOQ7Ngmx/IA==,type:str]
```
henrirosten: This adds an encrypted `something` secret, but I don't see any consumer for it. Is this leftover test data? If so, please remove it.

joinemm: oh good catch that indeed was test data

henrirosten (approved):

PR #1225, opened 2026-08-18
Fix network interface name on nethsm-gateway-dev

joinemm:
The /29 network was already defined, but the interface names differ between prod and dev gateways, so the configuration never took any effect.


ktusawrk (approved):
Cross-checked the interface name with `ip address` on the dev-gateway

PR #1227, opened 2026-08-18
Fix loki logging on uae hosts

joinemm:
I traced the incoming dropped packages on ghaf-monitoring to uae azureci, which was trying to send logs into nebula address of ghaf-monitoring on port 3100. This port is not open on the nebula firewall and as such the packets were being dropped. 

Fixed by using the public monitoring.vedenemo.dev endpoint with basic auth as we do on other hosts. As a bonus, the logs from uae are now actually present. They were never received on monitoring end before now.

Also changed nebula behaviour to instantly reject packets that hit the firewall, instead of dropping them and timing out. This will make things easier to debug in the future.

### DIFF
--- a/hosts/uae/common.nix
+++ b/hosts/uae/common.nix
@@ -4,11 +4,10 @@
   config,
   lib,
   self,
-  machines,
   ...
 }:
 let
-  defaultLoki = "http://${machines.ghaf-monitoring.nebula_ip}:3100";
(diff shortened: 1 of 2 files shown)


fayadf (approved):

fayadf:
Well .. we just explored the possibilities.

fayadf:
Thanks for the fix!

PR #1241, opened 2026-08-26
Implement SLSA 1.2 source VSA generator

joinemm:
Add a SLSA 1.2 Source VSA generator that:

- Issues keyless Sigstore VSA bundles for `main` branch commits
- Publishes bundles to GHCR
- Supports fetching, inspecting, and verifying bundles
- Validates workflow identity, source refs, commits, and policy
- Provides a GitHub Actions workflow and reusable composite action

## Manual setup

After first publication, make the `ghaf-infra/source-vsa` GHCR package public in its package settings (https://github.com/orgs/tiiuae/packages), then rerun the workflow.

A later hardening PR will follow once we have stable working action to pin to.

### DIFF
--- a/pkgs/source-vsa/src/source_vsa.py
+++ b/pkgs/source-vsa/src/source_vsa.py
@@ -45,6 +45,18 @@ def ref_allowed(ref: str, allowed_refs: list[str]) -> bool:
     return any(re.fullmatch(pattern, ref) for pattern in allowed_refs)
 
 
+def verifier_matches(verifier: Any, context: dict[str, Any]) -> bool:
+    """Match the verifier identity and an action revision when independently set."""
+    return (
+        isinstance(verifier, dict)
+        and verifier.get("id") == context["verifier_id"]
+        and (
+            "action_sha" not in context
+            or verifier.get("version") == {"action": context["action_sha"]}
+        )
+    )
+
+
 def run_command(
     arguments: list[str],
     *,
@@ -204,18 +216,6 @@ def statement_from_bundle(bundle_path: Path) -> dict[str, Any]:
     return statement
 
 
-def action_sha_from_bundle(bundle_path: Path) -> str:
-    """Read the composite-action version claimed by a VSA bundle."""
-    statement = statement_from_bundle(bundle_path)
-    try:
-        action_sha = statement["predicate"]["verifier"]["version"]["action"]
-    except (KeyError, TypeError) as error:
-        raise SourceVSAError("VSA issuer action version is missing") from error
-    if not isinstance(action_sha, str):
-        raise SourceVSAError("VSA issuer action version is invalid")
-    return action_sha
-
-
 def verify_statement(statement: dict[str, Any], context: dict[str, Any]) -> None:
     """Require the exact statement shape and policy decisions used by this issuer."""
     expected_subject = {
@@ -247,10 +247,6 @@ def verify_statement(statement: dict[str, Any], context: dict[str, Any]) -> None
         raise SourceVSAError("VSA sourceRefs are not allowed by policy")
 
     expected_predicate = {
-        "verifier": {
-            "id": context["verifier_id"],
-            "version": {"action": context["action_sha"]},
-        },
         "resourceUri": f"git+https://github.com/{context['repository']}",
         "verificationResult": "PASSED",
         "verifiedLevels": [context["verified_level"]],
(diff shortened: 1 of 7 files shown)


pkgs/source-vsa/src/source_vsa.py
```diff
@@ -0,0 +1,585 @@
...
+    """Pull a bundle and distinguish an absent tag from registry failures."""
+    arguments = ["oras", "pull", "--output", str(destination)]
+    if registry_config is not None:
+        arguments.extend(["--registry-config", str(registry_config)])
+    result = run_command([*arguments, reference], check=False)
+    if result.returncode:
+        detail = result.stderr.strip() or result.stdout.strip()
+        missing = any(
+            marker in detail.lower()
+            for marker in ("not found", "manifest unknown", "404")
```
henrirosten: This treats missing bundles by substring-matching `oras` stderr, which is fragile.

Authenticated GHCR can return `404 MANIFEST_UNKNOWN` for a missing package, but anonymous GHCR can return `denied: requested access to the resource is denied`. Please confirm the repo-scoped `GITHUB_TOKEN` gets the 404 form before first merge; otherwise first issuance aborts before `oras push`, the package is never created, and the "make it public and rerun" path is unreachable.

This should use a structured registry signal, such as manifest status / `MANIFEST_UNKNOWN`, rather than stderr substrings.

pkgs/source-vsa/src/source_vsa.py
```diff
@@ -0,0 +1,585 @@
...
+        )
+
+
+def verify(args: argparse.Namespace) -> None:
+    """Verify a local VSA bundle."""
+    context = policy_context(
+        load_policy(args.policy),
+        args.repository,
+        args.commit,
+        action_sha_from_bundle(args.bundle),
```
henrirosten: `verify()` derives the expected issuer action SHA from the same bundle it is about to verify. As a result, the `predicate.verifier.version.action` check compares the DSSE payload against itself and cannot reject a bundle genuinely signed by an untrusted revision of the issuer action.

`fetch()` has the same issue at line 510. Also, `action_sha_from_bundle()` parses the DSSE payload before `verify_bundle()` runs, so unverified input is used to build the policy context.

The expected action revision needs to come from policy, or from an explicit caller-supplied trusted pin, rather than from the artifact under verification.

pkgs/source-vsa/src/source_vsa.py
```diff
@@ -0,0 +1,585 @@
...
+                    f"org.opencontainers.image.source=https://github.com/{context['repository']}",
+                    "--annotation",
+                    f"dev.slsa.source.commit={context['commit']}",
+                    context["reference"],
+                    f"{BUNDLE_FILENAME}:{BUNDLE_MEDIA_TYPE}",
+                ],
+                cwd=workdir,
+            )
+        else:
+            verify_bundle(bundle, context)
```
henrirosten, line 515: For the normal release flow, a `ghaf-*` tag will usually point at a commit that was already attested when it landed on `main`. Since the OCI reference is keyed only by commit, the pull at lines 406-411 finds the existing bundle and this branch only re-verifies it.

Example: commit C lands on main, producing a VSA with `sourceRefs == ["refs/heads/main"]`. Later `ghaf-1.0` is created on C; the workflow runs, finds the commit-scoped bundle, and succeeds without creating any VSA whose `sourceRefs` contains `refs/tags/ghaf-1.0`.

That does not match the tag trigger/description. Either make the stored artifact ref-scoped so tag pushes produce their own bundle, or drop/document the tag trigger if these VSAs are intentionally commit-scoped only.

ktusawrk (dismissed):

pkgs/source-vsa/src/source_vsa.py
joinemm: resolved by removing explicit ghaf-* tags. main branch will generate VSA and that VSA can be used still the commit is tagged

pkgs/source-vsa/src/source_vsa.py
joinemm: fixed

pkgs/source-vsa/src/source_vsa.py
joinemm: fixed. self-derived check removed. future work should pin the action version sha to a stable version

henrirosten (approved):

PR #1243, opened 2026-08-27
Enable multi-tenancy in ghaf-log loki config

joinemm:
Enables the use of `tenant_id` in https://github.com/tiiuae/ghaf/pull/2172, to avoid hitting per-user limits.

something to note is that once we enable multi-tenancy on `ghaf-log` then sending any logs without tenant id present will fail

Also add testing team access to ghaf-log

### DIFF
--- a/hosts/ghaf-log/configuration.nix
+++ b/hosts/ghaf-log/configuration.nix
@@ -19,6 +19,7 @@
     openssh
     nginx
     team-devenv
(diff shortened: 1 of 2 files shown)


henrirosten (approved):

joinemm:
closing as multi-tenancy turned out to be a bad solution for this problem

PR #1258, opened 2026-09-07
Fix experimental features type for newer nixpkgs

henrirosten:
Fix flakevuln comparison scans failing with newer nixpkgs, which requires nix.settings.experimental-features to be a list. Update both the shared and Azure modules, preserving compatibility with the pinned nixpkgs.


henrirosten:
Closed in favor of: https://github.com/tiiuae/ghaf-infra/pull/1259

PR #1273, opened 2026-09-10
Package hetzci jenkins and testagent

joinemm:
Adds a generic Jenkins module with options to define all secrets and tii-specific variables, as well as similar module for defining a testagent.

Note that the actual pipelines are still full of hardcoded variables that refer to ghaf and our infra but that is out of scope for this PR. Signing and the pkcs11-proxy are also out of scope until it's clearer what we do with that.


henrirosten (commented):
`.github/workflows/update-flake-inputs.yml`: github action workflow for weekly plugin update still references the old Jenkins plugin path. This PR moves `hosts/hetzci/plugins.txt` and `hosts/hetzci/plugins.json` to `modules/jenkins/`, but `.github/workflows/update-flake-inputs.yml` still calls `update-jenkins-plugins hosts/hetzci` and still adds `hosts/hetzci/plugins.json` to the resulting PR.

modules/jenkins/default.nix
```diff
@@ -84,21 +87,54 @@ let
...
     envType = lib.mkOption {
-      type = lib.types.enum [
-        "dbg"
-        "dev"
-        "prod"
-        "release"
-        "vm"
-      ];
-      description = "The type of environment this is";
+      type = lib.types.str;
```
henrirosten: This relaxes `envType` from the previous enum (`dbg`, `dev`, `prod`, `release`, `vm`) to an unrestricted string.

The Jenkins pipelines branch on exact `CI_ENV` values for scheduling, signing, archive behavior, release handling, and VM behavior. A typo would now evaluate successfully but silently change runtime behavior.

modules/jenkins/default.nix
```diff
@@ -285,13 +298,113 @@ in
...
-    users.users.jenkins.homeMode = "710";
-    systemd.services.caddy.serviceConfig.SupplementaryGroups = [ "jenkins" ];
+    users.users =
+      lib.mapAttrs (_: publicKey: {
+        isNormalUser = true;
+        openssh.authorizedKeys.keys = [ publicKey ];
+      }) cfg.nodes.authorizedKeys
+      // lib.optionalAttrs config.services.caddy.enable {
+        jenkins.homeMode = "710";
+      };
```
henrirosten: This constructs `users.users` with the shallow `//` operator. If `nodes.authorizedKeys` ever contains a key named `jenkins`, this attrset merge would silently overwrite one side of the Jenkins user definition instead of letting the module system merge option definitions normally. Since this is now exported as a reusable module, please express these as mergeable module definitions or reject reserved usernames.

henrirosten (dismissed):

joinemm:
rebased to fix merge conflict

PR #1274, opened 2026-09-10
nix-fast-build.sh: Remove flake source archival workaround

cazfi:
nix-fast-build works correctly now, no need for the bug workaround.


alextserepov (approved):
Perhaps better if someone from Nix experts will take a look as well, but looks good to me.

henrirosten (approved):

PR #1276, opened 2026-09-14
Revert "Unify kernel versions"

henrirosten:
This reverts commit df695e8a214c40f2079f231720b9bebe4c8cb8aa.

I did not manage to `inv install hetzarm-dbg-1` without this revert, and also saw issues with installing hetzarm-release builder earlier. I think the commit needs more testing and analysis to understand why df695e8a21 causes kexec failure especially on those arm hosts, but meanwhile I think it needs to be reverted to unblock properly using the infra.


joinemm (approved):
probably we could consider using the latest kernel on all hosts

PR #1280, opened 2026-09-15
Enable custom build and orin flash

leivos-unikie:
Add parameter field for custom ghaf target to ghaf-manual pipeline.

Enable launching orin agx flash script in ghaf-manual. This part to work requires adding USB cable between test agent and target Orin AGX and downloading Linux_for_Tegra on the test agent.

For now this works only with flash-qspi. Running flash-script would require building also ghaf image in addition to flash script. Running flash-qspi with enabling secure boot would also require building the ghaf image in addition to flash-qspi script.

Demonstrated on dev:
- first manually downloaded Linux_for_Tegra driver package on dev test agent /var/lib/nvidia/
- connected USB-C to C cable from test agent to Orin AGX64
- It builded packages.x86_64-linux.nvidia-jetson-orin-agx64-debug-from-x86_64-flash-qspi and flashed orin-agx-64
https://ci-dev.vedenemo.dev/job/ghaf-manual/843/


ktusawrk (approved):
This PR enables also building custom Ghaf targets, which is a very desirable feature. Full feature set for flashing AGX devices requires additional functionality, like installing linux for Tegra drivers from Nixos configuration, but let's handle them separately.

PR #1281, opened 2026-09-16
Use ghaf-dev cache on ci-dbg builders

henrirosten:
Allow the ci-dbg builders to reuse artifacts from the ghaf-dev binary cache. If an identical derivation has already been built by ci-prod or ci-dev, it can be downloaded instead of rebuilt, reducing build times and duplicated compute usage in ci-dbg.

The builders still push their results to the ghaf-dbg cache, so the Cachix push path remains covered by ci-dbg testing. This also makes ghaf-dbg effectively disposable: its contents can be cleared at little cost because most reusable artifacts remain available from ghaf-dev, leaving only debug-specific artifacts to rebuild: this also still allows testing the builder configuration tweaks in ci-dbg builders.


joinemm (approved):

joinemm:
good change, building anything on dbg without the cache is impossible

PR #1286, opened 2026-09-17
Update inputs (Sep26 security update 2)

cazfi:

PR #1289, opened 2026-09-18
Add baseline reset and adopt it on ci-dbg

henrirosten:
Add a NixOS module that restores hosts from a read-only Btrfs baseline on every boot while preserving machine and SSH identity and enable it in ci-dbg hosts.

For a full description of the feature, see [baseline-reset.md](https://github.com/tiiuae/ghaf-infra/blob/374e505b018a0ded7b5b32a7c41c4fe945d731b5/docs/baseline-reset.md) added in this PR.

After this PR: enabling or disabling baseline reset changes the disk layout and requires `inv install` on the initial install after enrollment (currently, only impacts ci-dbg).

Tested extensively on ci-dbg: `inv install` round trips main --> baseline reset --> main on all three hosts; `deploy --boot` and `deploy` plus reboot over two reset cycles; a release-candidate build across both builders; an in-flight builder reboot with clean-store build recovery; test-agent reconnection; `cachix-push` lifecycle.

Not included: ci-release adoption will done in a separate PR after ci-dbg and when the next ghaf release cycle is ready.


ktusawrk (approved):
This is not a technical review, but definitely an interesting concept.

henrirosten:
Rebased and re-tested after https://github.com/tiiuae/ghaf-infra/pull/1286 was merged

hosts/builders/hetz86-dbg-1/disk-config.nix
```diff
@@ -13,7 +13,7 @@
         };
         ESP = {
           type = "EF00";
-          size = "256M";
```
joinemm, line 16: lets bump these straight to 1G

docs/baseline-reset.md
```diff
@@ -0,0 +1,136 @@
...
+-->
+
+# Baseline reset
+
+Baseline reset returns a compatible NixOS host to a clean deployed state on
+every boot while preserving normal NixOS deployment. Runtime changes to the
+root filesystem and Nix store, including service and local build state, are
+discarded.
+
+Deployment saves the system's Nix store closure as a read-only btrfs baseline.
```
joinemm: btrfs "baseline" is not a real thing and is only made up name for this module. I would prefer if the real names of the features were used (snapshot), or at least mentioned.

joinemm (changes requested):

hosts/builders/hetz86-dbg-1/disk-config.nix
henrirosten: Updated all three ci-dbg ESPs from 512 MiB to 1 GiB in d56580d9.

docs/baseline-reset.md
henrirosten: Updated the documentation in d56580d9 to use the underlying btrfs terminology.

joinemm (approved):

